PR #75, opened 2016-03-08
SB-230: Implement metadata merging for the ArtifactDeployer

baisjuan:


strongbox-client/src/main/java/org/carlspring/strongbox/client/ArtifactClient.java
```diff
@@ -410,12 +386,11 @@ public boolean pathExists(String path)
         setupAuthentication(resource);
 
         Response response = resource.request(MediaType.TEXT_PLAIN).get();
-
+        System.out.println(response.getStatus());
```
carlspring: I don't think we need this.

carlspring:
Thanks for all your hard work on this! :)

PR #83, opened 2016-03-15
SB-420: Spring Security authentication and authorization

martin-g:
This Pull Request adds authen/authorization with Spring Security.
`strongbox-security-context.xml` sets up two providers 
- one that uses custom implementation based on OrientDB (first time using OrientDB, so be kind! :-) )
- one that uses simple provider with hardcoded users in the context.xml

To add other types of providers see the links in the .xml file.

To add more security:
- add more `<security:intercept-url>` entries in the context.xml and define which roles are allowed/denied to reach them
- provide REST APIs to add/edit/remove users (with roles)

### DIFF
--- a/strongbox-parent/pom.xml
+++ b/strongbox-parent/pom.xml
@@ -174,6 +174,18 @@
             </dependency>
 
             <dependency>
+                <groupId>com.orientechnologies</groupId>
+                <artifactId>orientdb-core</artifactId>
+                <version>${version.orientdb}</version>
+            </dependency>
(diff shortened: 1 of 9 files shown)


martin-g:
At the moment there are two demo users:
- `martin/agent007` - an Admin user who is allowed to open `/configuration/strongbox/xml`
- `james/bond` - a normal (i.e. non-admin) user who is allowed to open any REST endpoint but `/configuration/strongbox/xml`

PR #87, opened 2016-03-23
#SB-462: Investigate and fix the issue with deep nested group repositories

baisjuan:
- Fixed fix in GroupLocationResolver
- Updated test cases in GroupLocationResolverTest
- Added test case in GroupLocationResolverTest to test the fix


carlspring:
After having a proper look at the code, I'm afraid I'm not quite convinced it works properly. 

Please, rename `searchArtifact` to `resolveArtifact`.

These sections are not returning properly:

                        Repository repository = getConfiguration().getStorage(sId).getRepository(rId);
                        if (repository.isInService() && basicRepositoryService.containsPath(repository, artifactPath))
                        {
                            final ArtifactInputStream is = searchArtifact(sId,repository,artifactPath);
                            // This doesn't return anything
                        }

carlspring:
Many thanks yet again! :)

PR #101, opened 2016-04-29
SB-513: Migrate Spring contexts and web.xml to Java configurations

ivanursul:
First step is to remove optional beans. They can be removed, because we have 
<context:component-scan = "org.carlspring.strongbox" />

Next step - intrtoducting java contexts and doing removing everything from xml


ivanursul:
Removed xml files and created fully java config application.
I would recommend to do separate package per module

ivanursul:
When you write tests - always see what dependencies do you need - if you have dependencies from other module - provide configuration for that module

carlspring:
Thanks for your great contribution and for all the hard work! :)

PR #107, opened 2016-05-04
Removed unused dependencies

ivanursul:
Because of some unused modules we have on strongbox, there was a need of some depependencies removal.


carlspring:
Thanks for fixing this as a follow-up of SB-520.

PR #108, opened 2016-05-04
Package per module

ivanursul:
I started with common module.


carlspring:
@ivanursul,

Thanks for this, but let's keep this as it is for the time-being. It's not optimal, I know... but if it starts getting unmanageable, I propose to then look into it. If it comes to this, a lot of the currently existing modules will have to be refactored, beyond just having their package names renamed, because we'll otherwise have so many packages contains things like `util` or `services`...

PR #109, opened 2016-05-05
Removed redundant context load for web-core module

ivanursul:


carlspring:
Thanks for fixing this! :)

PR #114, opened 2016-05-13
Resolve the exception below which happens when you build the strongbox-web-core 

AlexOreshkevich:
Hi Martin,

I think I've resolved issue with NoClassDefFoundError. Spending a Lot Of Time I found that we don't need jetty stop manual execution specified in pom.xml file, because we are using jetty as a deamon and as a part of our current maven build. When the build ends jetty will be stopped automatically and all resources will be released.

As a proof I was inspected process list at my machine, there was no separate process for jetty, and there was no java processes after maven build was finished.

Please let me know if my fix works for you and if you are pleased with that solution.


carlspring:
Many thanks for your fix!

PR #144, opened 2016-06-30
SB-574 and SB-126 [1/2]

AlexOreshkevich:


AlexOreshkevich:
Hi Martin,

please review current state of the fork if you will had a chance. A few questions:

1) Do we need to store data from XLM config that relates to SB-126 in the database? How this will impact on the current user schema?
2) Do we need to allow define roles in the XML file with the same names as we have already predefined in java code? Do we need to merge them? 
3) How we need to process privileges without roles in XML config file? 
4) Do we need to process authorization XML file FIRST, and then user XLM file second, and validate second somehow? Depending on first config content?

carlspring:
Hi,

Thanks for all these good questions. I have to apologize here and admit I didn't initially have the clear idea and understanding I do now, after our discussions and seeing your work.

Regarding 1) and 2): Privileges should have a type - `built-in`, or `custom`.
a) The **built-in privileges should not be** importable/exportable via XML.
b) The **built-in privileges should be** populated in the database, upon the very first run of the server (after an installation), if they aren't in the database, but it should not be possible to modify them.
c) If the property for the XML file has been specified (question: did we actually define one), then the privileges and roles from the file will need to be imported to the database.

3) There could be unassigned/unused privileges. This is quite often the case and we should support it. Quite often there are just two types of users: admins and developers and there is no fine-graining of the security, because there is either no need, or there is a lack of knowledge among the people administering such servers of how to do it. Therefore, this has to be possible.

4) The validation should be against the database, not the XML file.

strongbox-security-api/src/main/java/org/carlspring/strongbox/security/jaas/Role.java
```diff
@@ -52,6 +54,21 @@ public Role(String name,
         this.description = description;
     }
 
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o) return true;
```
carlspring: Could you please always use curly braces around the bodies of `if` statements, even if they're one-liners?

strongbox-security-api/src/main/java/org/carlspring/strongbox/security/jaas/Roles.java
```diff
@@ -38,4 +38,12 @@ public void setRoles(Set<Role> roles)
         this.roles = roles;
     }
 
+    @Override
+    public String toString()
+    {
+        final StringBuilder sb = new StringBuilder("\n\tRoles{");
+        sb.append("roles=").append(roles);
+        sb.append('}');
+        return sb.toString();
```
carlspring: Another code-style one, could you please always put an empty line above the `return` statement? It won't make the code better, or anything, but it will make it more obvious where the return statement is. (I've fixed most of the other cases).

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/configuration/ConfigurationRepository.java
```diff
@@ -1,144 +1,166 @@
...
+@Transactional
 public class ConfigurationRepository
 {
 
     private static final Logger logger = LoggerFactory.getLogger(ConfigurationRepository.class);
-
     @Autowired
-    OObjectDatabaseTx db;
+    ServerConfigurationService serverConfigurationService;
+    @Autowired
```
carlspring: (Just another stylistic one): Could you please leave empty lines between the definitions of fields?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/Storage.java
```diff
@@ -3,37 +3,42 @@
...
 public class Storage
+        implements Serializable
 {
 
+    /**
+     * Added to avoid a runtime error whereby the detachAll property is checked for existence but not actually used.
+     */
+    @JsonIgnore
+    protected String detachAll;
+    @Version
```
carlspring: Could you please leave empty lines between fields? Thanks.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/Repository.java
```diff
@@ -3,28 +3,34 @@
...
 public class Repository
+        implements Serializable
 {
 
+    /**
+     * Added to avoid a runtime error whereby the detachAll property is checked for existence but not actually used.
+     */
+    @JsonIgnore
+    protected String detachAll;
+    @Version
```
carlspring: Could you please leave empty lines between fields?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/Repository.java
```diff
@@ -359,4 +344,75 @@ public void setHttpConnectionPool(HttpConnectionPool httpConnectionPool)
...
+    {
+        return version;
+    }
+
+    public void setVersion(Long version)
+    {
+        this.version = version;
+    }
+
+    public boolean isAllowsForceDeletion()
```
carlspring: @AlexOreshkevich,

Could you please restore the original method names here (and below)? 

* `isAllowsForceDeletion()` --> `allowsForceDeletion()`
* `isAllowsRedeployment()` --> `allowsRedeployment()`
* etc, etc....

AlexOreshkevich: We can’t do it that way because such kind of names breaks JavaBeans Naming Conventions that’s are part of java.io.Serializable requirements that needed to store entities in the OrientDB properly
(of course, keeping in mind current simplifications with BinaryConfiguration)

strongbox-security-api/src/main/java/org/carlspring/strongbox/security/jaas/Roles.java
AlexOreshkevich: This rule (empty line before return statement) is missing in XML template that I was imported into IDE and need to be implemented manually (that could be annoing).

Could you modify code style template as well? 

PS I will push that empty line for that case.

PR #145, opened 2016-07-01
SB-600 Sort out the jackson issues.

korest:
Fixed now it should work as expected with jackson


carlspring:
Thanks a lot for this fix! ;)

PR #170, opened 2016-10-14
SB-575 [merged with carlspring:master]: Replace the REST-related test cases with rest-assured

carlspring:


strongbox-web-core/src/test/java/org/carlspring/strongbox/rest/ArtifactControllerTest.java
```diff
@@ -39,66 +36,49 @@
...
 
-    @Before
-    public void setUpClass()
+    @BeforeClass
+    public static void setUpClass()
             throws Exception
     {
+        // to make this test idempotent (in general) we will check and remove previous generated artifacts if they are present
+        // notice that we can't delete the whole repository basedir because it contains also .index, .trash etc.
+        removeDir(REPOSITORY_BASEDIR_RELEASES.getAbsolutePath() + "/org");
```
carlspring: @AlexOreshkevich ,

I understand what you mean, but this is not a good idea. My reasoning is this: at some point in the very near future, I would like to be able to run the tests in parallel (see [SB-320](https://dev.carlspring.org/youtrack/issue/SB-320). If you delete these directories each time the `setUpClass` is invoked before each test, then some of the running test methods might fail, if they still need the resources.

I've been very particular and put specific comments next to each `generateArtifact` statement so that it is absolutely clear which test method is using the respective generated artifact. None of the test's methods are using each other's resources in any way.

strongbox-web-core/src/test/java/org/carlspring/strongbox/rest/common/RestAssuredBaseTest.java
```diff
@@ -0,0 +1,255 @@
...
+    protected void delete(String storageId,
+                          String repositoryId,
+                          String path,
+                          boolean force)
+            throws ArtifactOperationException
+    {
+        String url = getContextBaseUrl() + "/storages/" + storageId + "/" + repositoryId;
+
+        given().contentType(MediaType.TEXT_PLAIN_VALUE)
+               .param("path", path)
```
carlspring: @AlexOreshkevich,

Unless I'm mistaken, this variable needs to go, as it's all path-variable based, and not using the request parameters.

AlexOreshkevich:
Unit tests should be independent. If you generate something somewhere, consume it at that place and don’t reuse somewhere else.

You have initial mistake for that generation and Im just trying to provide simple and quick workaround to make it working properly.

carlspring:
Thanks for all your efforts to get this implemented, stabilized and merged!

PR #177, opened 2016-11-15
SB-671: Implement Nuget repositories

carlspring:


strongbox-resources/strongbox-storage-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.xml
```diff
@@ -6,6 +6,11 @@
     <port>48080</port>
 
     <storages>
+        <storage id="nuget-common-storage">
+            <repositories>
+                <repository id="releases" policy="release" implementation="file-system" layout="Nuget Hierachlical" type="hosted" allows-force-deletion="true" checksum-headers-enabled="true"/>
```
carlspring: There is a typo here:

     <repository id="releases" policy="release" implementation="file-system" layout="Nuget Hierarchical" type="hosted" allows-force-deletion="true" checksum-headers-enabled="true"/>

(not `Hierachlical`).

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/NugetArtifactCoordinates.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public String getVersion()
+    {
+        return getCoordinate(VERSION);
+    }
+
+    @Override
+    public void setVersion(
```
carlspring: @sbespalov , 

This is just a cosmetic remark: could please format first parameter of the methods to be on the same line as the method and then align all other parameters, under the first character?

For example:

    public void setBlaBla(String version,
                          String foo)

Could fix this across the rest of the code as well, please? Thanks! :)

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/NugetHierarchicalArtifactCoordinates.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public String toPath()
+    {
+        String idLocal = getId();
+        String versionLocal = getVersion();
+        return String.format("%s/%s/%s.%s.nupkg", idLocal, versionLocal, idLocal, versionLocal);
+    }
+
+    public static void main(
```
carlspring: Could you please extract this into a separate test called `NugetHierarchicalArtifactCoordinatesTest` instead of using a `main` method?

strongbox-parent/pom.xml
```diff
@@ -583,6 +583,39 @@
             </releases>
         </repository>
     </repositories>
+    <pluginRepositories>
```
carlspring: As discussed, this is already handled by the `settings.xml` file I showed you. You should be using that instead of defining it in the `pom.xml`-s. That's how we have it across the rest of the project.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/Maven2LayoutProvider.java
```diff
@@ -46,12 +46,6 @@
     public static final String ALIAS = "Maven 2";
 
     @Autowired
```
carlspring: :+1: Thanks for this cleanup! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryLayoutEnum.java
```diff
@@ -14,7 +14,9 @@
 
     RPM("RPM"),         // Unsupported
 
-    YUM("YUM");         // Unsupported
+    YUM("YUM"),         // Unsupported
+    
+    NUGET_HIERACHLICAL("Nuget Hierachlical");
```
carlspring: Same typo. :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/web/HeaderMappingFilter.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author Sergey Bespalov
+ *
+ */
+@WebFilter(filterName = "headerMappingFilter", urlPatterns = "/storages/*")
+public class HeaderMappingFilter implements Filter
+{
+
+    private static final String USER_AGENT_UNKNOWN = "unknown";
+    private static final String USER_AGENT_NUGET = "NuGet";
+    private static final String USER_AGENT_MAVEN = "Mavent";
```
carlspring: There's a typo here as well -- `Mavent`, instead of `Maven`.

sbespalov:
all above done

strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/ArtifactController.java
```diff
@@ -27,32 +17,50 @@
...
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
 
 /**
  * test
  */
 @RestController
-@RequestMapping("/storages")
+@RequestMapping(path = ArtifactController.ROOT_CONTEXT, headers = "user-agent=Maven/*")
 public class ArtifactController
```
sbespalov: I think we can rename it after pull request will be merged, to prevent conflicts. As about separate controller per layout provider I think it is better to say that we have controllers per protocol (Maven, Nuget, Docker etc.)

sbespalov: I think we must discuss one question about storage-api interfaces. Seems that we should use 'ArtifactCoordinates' instead of 'path' in parameters.
For example this 'void org.carlspring.strongbox.services.ArtifactManagementService.store(String storageId, String repositoryId, String path, InputStream is)' method must look like this 'Open Declarationvoid org.carlspring.strongbox.services.ArtifactManagementService.store(String storageId, String repositoryId, ArtifactCoordinates path, InputStream is)'


hope you understand my suggestion.
I can try to explain it more carefully.

carlspring:
Hi,

I don't think I get why this is needed? ArtifactCoordinates can translate
to a path via the toPath() method.

Would you mind further explaining why this is required?

Kind regards,

Martin

strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/ArtifactController.java
carlspring: Hi,

I don't think I get why this is needed? ArtifactCoordinates can translate
to a path via the toPath() method.

Would you mind further explaining why this is required?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryLayoutEnum.java
```diff
@@ -14,7 +14,9 @@
 
     RPM("RPM"),         // Unsupported
 
-    YUM("YUM");         // Unsupported
+    YUM("YUM"),         // Unsupported
+    
+    NUGET_HIERACHLICAL("Nuget Hierarchical");
```
carlspring: I think this class should actually just get deprecated. It's used only in the [Repository](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/Repository.java).

[This line](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/repository/Repository.java#L47):

    private String layout = RepositoryLayoutEnum.MAVEN_2.getLayout();

Should now be changed to just:

    private String layout = Maven2LayoutProvider.ALIAS;

And the whole `RepositoryLayoutEnum` could go, making things much more dynamic... I mean... When someone else decides to add a new `LayoutProvider`, they won't have to change this class and so on.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/ArtifactController.java
sbespalov: Earlier we decided that each `Protocol` (Maven, Nuget, etc.) can have multiple layouts (Hierarchical, Plane etc.) which is controlled by `LayoutProvider`, and `path` means the layout path,  that's from one side. From the other side, when we want to store artifact (package) in `Controller`, we don't know which `Layout` we have under `Storage` and `Repository` and we can't build its `Path`. So we need some kind of abstraction which will point the artifact in `Repository` and `ArtifactCoordinates` can be such type of abstraction. But you are right about `toPath()` method and in this case we must move toPath() method from `ArtifactCoordinates` to `LayoutProvider.toPath(ArtifactCoordinates)`.

sbespalov: The other variant of the `Path abstraction` is to use custom URI in the form of `srtongbox://storageId/repoId?artifactNameAndVersionQuery` (for Maven it can be `strongbox://storage0/releases?group=GroupId&artifact=ArtifactId&version=VersionNumber` and for NuGet `strongbox://storage0/releases?package=PackageName&version=VersionNumber`).

carlspring: The problem with the URI is that you don't know whether, or not the interacting tools will be able to figure out what to do with this "protocol".

Apart from that, I've been thinking: if each controller will have it's own supported `User-Agent`-s, then we have the following problem...

Maven 2 repositories could be used by many tools (both such that we know - Ant+Ivy, gradle, SBT, and custom ones). Apart from that, it would have to be possible (when there is a user interface in place) to deploy artifacts via the UI, meaning that we'd also have to support the browser's UA-s. This means, we'll have multiple UA-s per controller and it would have to be possible to manage UA-s per layout via a service class of some sort.

Would the controllers be able to handle this?

On your `*ArtifactController` question: you can always check what the layout is, by looking up the `Storage` and `Repository` from the `Configuration` and then doing a `repository.getLayout()`, so I'm still not quite sure I understand the problem...

sbespalov: >>>> Apart from that, I've been thinking: if each controller will have it's own supported User-Agent-s, then we have the following problem...

This seems not a problem because `Controllers` can handle multiply UA header values with `HeaderMappingFilter`. This filter in theory can map any HTTP context to `user-agent: Maven/*` or `user-agent: NuGet/*` for example.

sbespalov: > On your *ArtifactController question: you can always check what the layout is, by looking up the Storage and Repository from the Configuration and then doing a repository.getLayout(), so I'm still not quite sure I understand the problem...

We can use `Configuration` but  this is not so good by design in `Controller` level. Such an approach violates the abstraction principle of LayoutProvider.

> The problem with the URI is that you don't know whether, or not the interacting tools will be able to figure out what to do with this "protocol".

Interacting tools will not know anything about our internal URI format, they will use Controllers to interact with strongbox.

carlspring: My question here is: if both controllers are mapped to also support `Chrome`, `IE`, and `FF`, will this cause confusion? For example, if the accepted UA-s for the Nuget controller are `Nuget/*` and `Chrome/*` and the ones for the Maven controller are `Maven/*`, `Ivy/*`, ... `Chrome/*` will they not have conflicts?

sbespalov: > My question here is: if both controllers are mapped to also support Chrome, IE, and FF, will this cause confusion? For example, if the accepted UA-s for the Nuget controller are Nuget/* and Chrome/* and the ones for the Maven controller are Maven/*, Ivy/*, ... Chrome/* will they not have conflicts?

It will be conflict in this case, but I think that in this case we must use separate `Controller` for browsers.

carlspring: Also, I think that if we'll have different controllers, then we need to extract the code from the current `ArtifactController` into a `BaseArtifactController`, which then the `MavenArtifactController` and `NugetArtifactController` would extend. I don't know, if we really need to re-implement methods like `copy(...)`, `delete*()` and so on for each layout?

carlspring: If we have a separate controller for the browsers, then this will all go mad! So many controllers for the same thing...?

sbespalov: >Also, I think that if we'll have different controllers, then we need to extract the code from the current ArtifactController into a BaseArtifactController, which then the MavenArtifactController and NugetArtifactController would extend. I don't know, if we really need to re-implement methods like copy(...), delete*() and so on for each layout?

I also thought about this. We can make some refactoring after Nuget Controller will be completed and we will see which logic is the same and can be extracted somewhere.

carlspring: My advice is this: implement the Nuget repository support first and let's make sure things work properly. Once this is done, we'll have to outline the issues, as well as the similarities between these controllers and then carry out a refactoring. We should get @nenko-tabakov involved as well, because he's working on the implementation of OSGi repoistories (P2, and OBR)!

sbespalov: > If we have a separate controller for the browsers, then this will all go mad! So many controllers for the same thing...?

This will be not the same things. Browser controller will have different logic and methods which will differs from Maven and Nuget Controllers. For example you can not push packages with browser, you can only browse or search packages there. So I think we will need a separate Browser controller in any case.

carlspring: This is not exactly true:
> 
For example you can not push packages with browser, you can only browse or search packages there.

It should be possible to also upload artifacts manually -- via browser.

sbespalov: >It should be possible to also upload artifacts manually -- via browser. Most of the other repository managers support this.

You are right, but this is not the same with `nuget push` command and most likely we can't use `NugetController` methods for it.

sbespalov: In case of artifact upload in Browser it will be done with HTML form and we will have a separate controller to handle this form submition.

carlspring: Okay, I see what you mean.

carlspring:
@cs-devops-bot rebuild

carlspring:
Closing as agreed.

PR #185, opened 2016-11-23
SB-668 Implement a proper path-variable URL converter for Spring MVC

AlexOreshkevich:


strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/AuthorizationConfigController.java
```diff
@@ -192,8 +192,8 @@ public ResponseEntity deleteRole(@ApiParam(value = "The name of the role", requi
     // Assign privileges to the anonymous user
 
     @RequestMapping(value = "anonymous/privileges",
-                    method = RequestMethod.POST,
-                    consumes = MediaType.APPLICATION_JSON_VALUE)
+            method = RequestMethod.POST,
```
carlspring: **Trivial/Cosmetic remarks:**

- There seems to be no change here and below (unless I'm not seeing it)...?  Could you please restore the old formatting?
- Also, method parameters should always be aligned to the first parameter like this:

      @RequestMapping(value = "anonymous/roles",
                      method = RequestMethod.POST,
                      consumes = MediaType.APPLICATION_JSON_VALUE)

(Both in methods and in annotations)

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
```diff
@@ -53,8 +65,26 @@ public void configureMessageConverters(List<HttpMessageConverter<?>> converters)
...
+
+    @Override
+    public void configurePathMatch(PathMatchConfigurer configurer)
+    {
+        configurer
+                .setUseSuffixPatternMatch(true)
+                .setUseTrailingSlashMatch(false)
+                .setUseRegisteredSuffixPatternMatch(true)
+                .setPathMatcher(antPathMatcher);
+    }
```
carlspring: Cosmetic/Code style:

For all similar places, could you please align the methods based on the first one's starting index like this?

       configurer.setUseSuffixPatternMatch(true)
                 .setUseTrailingSlashMatch(false)
                 .setUseRegisteredSuffixPatternMatch(true)
                 .setPathMatcher(antPathMatcher);

carlspring:
@AlexOreshkevich ,

Your code looks great, (I haven't tried it, but I trust it behaves as expected)! Thanks for your hard work!

I have just one big remark: please, don't reformat code, unless necessary... Going over your changes, there are a lot of lines which have been reformatted, that have no actual change. It's really hard, (or rather -- time consuming) to read a diff like this and figure out what is part of the change and what is not. Don't reformat the entire class you're working on. There is no point. I've most-likely already gone and changed the formatting.

In this regard, could I ask you to please go over your changes in the pull, and restore the original formatting? I've also made a couple of formatting-related comments. I know -- we all have styles and this is not mission critical, but we already have a convention across the codebase and I think the code is tidy and easy to read.

Thanks! :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
AlexOreshkevich: Agreed

strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/AuthorizationConfigController.java
carlspring: How about if you move the exclusion above the `.idea/` and then commit just the `.gitignore` and after that commit the exluded xml file?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controller/ArtifactController.java
```diff
@@ -73,11 +73,12 @@ public ResponseEntity greet()
     @ApiResponses(value = { @ApiResponse(code = 200, message = "The artifact was deployed successfully."),
                             @ApiResponse(code = 400, message = "An error occurred.") })
     @PreAuthorize("hasAuthority('ARTIFACTS_DEPLOY')")
-    @RequestMapping(value = "{storageId}/{repositoryId}/**", method = RequestMethod.PUT)
+    @RequestMapping(value = "{storageId}/{repositoryId}/{subPath:.+}", method = RequestMethod.PUT)
```
carlspring: Okay,  it's great that you can specify any path variable name! :)
It's just that everywhete else it's called `path`... Even across the service and util classes and this way it will be more consistent across the codebas.

AlexOreshkevich:
The reason  for formatting issues is the following.

For historical reasons I personally use coding style where curly braces lies on the same line. But from project to project coding styles differs, and I simply use predefined IDE formatting based on codeStylesSettings.xml. Any file Im touching I need to reformat according to project code style settings, it's my general rule. Just to make sure that I don't violate any project-specific formatting rule. It could be hard or even impossible to change that habit.

carlspring:
Right... I see. Let's try to sort out the `.idea/codeStyle.xml` then and control it better through there. It's all up to personal habits and preferences, I know, and it's a "stupid" requirement, but when you have a diff of 150 lines and the actual change is 35 lines, I guess you know what I mean... Hehe...

Did you see my comment regarding the `.gitignore` file?

carlspring:
Thanks for your great contribution! :)

PR #183, opened 2016-11-21
SB-710: Make it possible to specify custom configuration for repositories

carlspring:
Implemented.

PR #186, opened 2016-11-23
Fixed ObjectFactories and added deserialization test.

highsource:
As discussed in this [SO](http://stackoverflow.com/questions/40707661/jaxb-inheritance-dymanically-named-tags/40719657?noredirect=1#comment68751856_40719657) post.

I get the following XML:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<repository id="test-repository" policy="mixed" implementation="file-system" layout="Maven 2" type="hosted" secured="false" status="In Service" trash-enabled="false" indexing-enabled="false" allows-force-deletion="false" allows-deployment="true" allows-redeployment="false" allows-delete="true" allows-directory-browsing="true" checksum-headers-enabled="false">
    <aws-configuration bucket="test-bucket" key="test-key"/>
    <google-cloud-configuration bucket="test-bucket" key="test-key"/>
    <group/>
</repository>
```

It also cleanly unmarshals.


steve-todorov:
@cs-devops-bot add to whitelist @cs-devops-bot build

carlspring:
Hi Alexey!

Thanks for taking a look at it and giving it a go!

I'm afraid this still doesn't quite work. I had already gotten it to that stage. The problem is that the test you tried is the simpler one, which just serializes a `Repository` into `<repository .../>`. This indeed works fine until it has to enter the bigger picture -- when `Repository` is used in the real-life scenario. Things should look something like this:

    <configuration>
        <storages>
            <storage id="storage0">
                <repositories>
                    <repository id="test-repository" ...>
                        <aws-configuration bucket="test-bucket" key="test-key"/>
                        <google-cloud-configuration bucket="test-bucket" key="test-key"/>
                        <group/>
                    </repository>
                </repositories>
            </storage>
        </storages>
    </configuration>

What _it is_ currently printing instead is this:

    <configuration>
        <storages>
            <storage id="storage0">
                <repositories>
                    <repository id="test-repository" ...>
                        <custom-configuration/>
                        <custom-configuration/>
                        <group/>
                    </repository>
                </repositories>
            </storage>
        </storages>
    </configuration>

So, it's printing out two empty `<custom-configuration/>`-s.

Any ideas how this can be fixed?

Once again, many thanks for your time and help!

highsource:
You should be a little bit more specific about "real-life scenario". Do you want to marshal/unmarshal without specifying custom configuration classes *at all*? Because this is what I see here:

https://github.com/highsource/strongbox/blob/SB-710/strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/RepositoryTest.java#L73

carlspring:
Good catch! Seems like I've forgotten to pass `CustomConfigation.class` as a parameter to the `GenericParser`'s constructor.

carlspring:
@highsource ,

I tried it like this:

    GenericParser<Configuration> parser = new GenericParser<>(Configuration.class, CustomConfiguration.class);

But this doesn't seem to have any effect. Do you see anything else that could be wrong?

highsource:
Your JAXBContext should know all your custom configuration classes.

carlspring:
But the thing is -- in the test case, they're there for the purpose of testing. However, in the "real-life" scenario, these will be coming via "plugin" jars, so we won't know their names in advance in order to list them. So... is there a way to do this sorcery somehow?

highsource:
This is why I wrote you should describe what you mean by "real-life". Really, how could I have guessed about "plugin" jars?

The point is, JAXB *needs* to know this classes otherwise it won't unmarshal/marshal them. You have to feed these classes or packages to the JAXBContext, period. So you'll have to implement some kind of autodiscovery solution. One of the possible solutions to have some `META-INF/com.acme.foo.CustomConfiguration` resource in each plugin, listing custom configuration classes. Then you could gather all these `META-INF/com.acme.foo.CustomConfiguration` resources from all JARs (trivial via `ClassLoader.getResources(...)`, load and register classes from there.

There are surely mor options for the autodiscovery, but in any case you will need the classes or names of packages containing `ObjectFactory` classes.

carlspring:
@highsource,

Many thanks for your advice and help! I have now implemented this according to your suggestion using a service loader and gotten it to work in #183.

I will close this pull request.

Thanks again and all the best!

Martin

PR #193, opened 2016-12-04
SB-711: Implement artifact ArtifactEntry CRUD

carlspring:


strongbox-data-service/src/main/java/org/carlspring/strongbox/data/domain/GenericEntity.java
```diff
@@ -20,7 +20,7 @@
 {
 
     @Id
-    protected String id;
+    protected String databaseId;
```
carlspring: @AlexOreshkevich,

I don't think you've reverted the changes we discussed, regarding this. Or maybe you haven't pushed...?

carlspring: This should be changed to `objectId`, as we discussed.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/repository/ArtifactRepository.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author Martin Todorov
+ */
+@Transactional
+public interface ArtifactRepository
+        extends OrientRepository<ArtifactEntry>
+{
+
+    @Query("select * from ArtifactEntry where artifactCoordinates.groupId = ?")
+        // "select * from ArtifactEntry where artifactCoordinates.groupId = 'org.carlspring.strongbox'"
+    List<ArtifactEntry> findByArtifactCoordinates(String groupId);
```
carlspring: This doesn't look right...(?!)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactEntryService.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Transactional
+public interface ArtifactEntryService
+        extends CrudService<ArtifactEntry, String>
+{
+
+    /**
+     * Returns list of artifacts that matches search query defined as {@link ArtifactCoordinates} fields.
+     * By default all fields are optional and combined using logical AND operator.
+     * If all coordinates aren't present this query will delegate request to {@link #findAll()}.
```
carlspring: What does that mean? Could you give an example? If the coordinates have five parameters (`groupId`, `artifactId`, `version`, `type`, `packaging`) and one hasn't been specified, then it will do a `findAll()`?

AlexOreshkevich: It will do a query on all coordinates that present and have non-null values.

strongbox-resources/strongbox-common-resources/src/main/resources/logback.xml
```diff
@@ -38,10 +38,6 @@
         <level value="ERROR"/>
     </logger>
```
carlspring: @AlexOreshkevich ,

I think the logging will have to be restored as it was.

strongbox-commons/pom.xml
```diff
@@ -65,6 +65,13 @@
     </build>
 
     <dependencies>
+
+        <dependency>
```
carlspring: @AlexOreshkevich,

Why does the `strongbox-commons` module need to depend on the `strongbox-data-service`?

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/MavenArtifactCoordinates.java
```diff
@@ -12,9 +12,9 @@
 public class MavenArtifactCoordinates extends AbstractArtifactCoordinates
 {
 
-    private static final String GROUPID = "groupId";
+    private static final String GROUP_ID = "groupId";
```
carlspring: Could you please keep the old names of these variables? I know this doesn't comply with the convention and Sonar is probably complaining. Feel free to ignore this complaint in this case.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactEntryService.java
carlspring: Great! Thanks! :)

strongbox-commons/pom.xml
AlexOreshkevich: I can't remember the exact reason but we had to resolve something related to circular dependencies. I believe we still need that change strongly.

PR #195, opened 2016-12-08
Resolve path variable value parsing issue

AlexOreshkevich:


carlspring:
@sbespalov : Thanks for spotting and reporting this!
@AlexOreshkevich : Thanks for fixing this!

PR #203, opened 2017-01-09
SB-713; SB-715

sbespalov:
SB-713: `StorageProvider` refactored to use `java.nio.file.Path` abstraction
SB-715: Common methods for the LayoutProvider-s moved to the AbstractLayoutProvider


carlspring:
Thanks for your help! :)

PR #213, opened 2017-02-08
SB-674 Reuse embedded database for every module during testing

AlexOreshkevich:


carlspring:
Thanks for fixing this! :)

PR #214, opened 2017-02-09
SB-739 Upgrade OrientDB to 2.2.x

AlexOreshkevich:


strongbox-data-service/src/main/java/com/orientechnologies/orient/core/version/OVersionFactory.java
```diff
@@ -0,0 +1,68 @@
...
+
+package com.orientechnologies.orient.core.version;
+
+import com.orientechnologies.orient.core.serialization.OBinaryProtocol;
+
+/**
+ * @author Artem Orobets (enisher-at-gmail.com)
+ * @since 10/25/12
+ */
+public class OVersionFactory
```
sbespalov: This, and above, looks like a copy paste.
Can we use original classes?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/configuration/ProxyConfiguration.java
```diff
@@ -25,10 +29,6 @@
     @JsonIgnore
     protected String detachAll;
 
-    @Version
-    @JsonIgnore
-    protected Long version;
```
sbespalov: Shouldn't we use `GenericEntity` here (and below)?

PR #215, opened 2017-02-17
Resolve validation issue with Bean Validation provider

AlexOreshkevich:


strongbox-parent/pom.xml
```diff
@@ -416,9 +418,51 @@
...
+                <groupId>javax.enterprise</groupId>
+                <artifactId>cdi-api</artifactId>
+                <version>${version.cdi-api}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.validation</groupId>
+                <artifactId>validation-api</artifactId>
+                <version>${version.validation-api}</version>
+            </dependency>
```
AlexOreshkevich: We will definitely need validation itself. Hibernate validation implementation is kind of reference implementation and the best of all existing at the moment.

strongbox-user-management/src/main/java/org/carlspring/strongbox/users/security/AuthorizationConfigProvider.java
```diff
@@ -99,7 +98,7 @@ public void init()
             catch (OSerializationException e)
             {
                 config = null;
-                logger.error("Unable to reuse existing authorization config", e);
+                logger.error("Unable to renuse existing authorization config", e);
```
AlexOreshkevich: Kind of probably

strongbox-parent/pom.xml
AlexOreshkevich: Answering on your question directly, I can't find easy way of removing validation at all, but I believe that we will need it and it will be useful to have it in our project. So I even stopped to look at any options how to exclude validation api from our dependencies.

carlspring:
Closing this pull, as the issue no longer appears in the master.

PR #218, opened 2017-02-27
SB-742: Start using the org.owasp:dependency-check-maven plugin to scan the dependencies for vulnerabilities

carlspring:


carlspring:
@cs-devops-bot rebuild

PR #219, opened 2017-02-27
SB-740:  Implement a cron task for regenerating checksums (in Nuget repositories) 

kate-novik:


strongbox-commons/src/main/java/org/carlspring/strongbox/io/filters/MetadataFilenameFilter.java
```diff
@@ -0,0 +1,23 @@
+package org.carlspring.strongbox.io.filters;
+
+import java.io.File;
+import java.io.FilenameFilter;
+
+/**
+ * @author Kate Novik.
+ */
+public class MetadataFilenameFilter
```
carlspring: Could you please create separate classes for Maven and Nuget? At the moment this is something small, but the problem is that at some point we'll have support for Docker, NPM, Bower, YUM and so on and so on as repositories. As you can imagine, each of them will have their own formats for metadata...
Perhaps the respective `LayoutProvider` could return an instance of this filter...?

carlspring:
Handled on a different branch.

PR #220, opened 2017-03-03
SB-744: Extract the Maven 2.x layout provider code into a separate module

carlspring:
- Renamed the strongbox-storage-indexing module to strongbox-storage-maven-layout-provider.
- Extracted the Maven 2.0 layout provider code into the strongbox-storage-maven-layout-provider.

@kate-novik: I'm really sorry for breaking your code, but I'm doing this so that it's clear which code is Maven-layout specific and in order to avoid future confusion. I will also do the same for the Nuget code. This is the only sane way I can think of that would make it clear which code is layout-specific and which is not. There'll hopefully be more providers soon, so this need to be as little confusing, as possible.


carlspring:
@cs-devops-bot rebuild

PR #227, opened 2017-03-24
SB-617: Heavy refactoring of the strongbox-storage-* modules (Phase 2)

carlspring:
SB-617: Heavy refactoring of the strongbox-storage-* modules
------------------------------------------------------------------------
SB-759: Create an implementation for repository features

- Implemented.

SB-760: Implement repository creation strategies per layout provider

- Implemented.

PR #229, opened 2017-04-08
SB-603 Implement per repository access

AlexOreshkevich:
Implementation notes:

- If user has no general privilege for some action but features defined for him under security-users.xml config file, dynamic voting system will try to grant access for him
- During first authentication phase (related to Spring Filters) request URL will be pre-saved and stored under principal reference
- During second authentication phase (related to Controller method PreAuthorize rules) this URL will be used for repositoryId extraction and lookup for additional privileges for current principal through Features reference

Possible improvements:

- repositoryId parsing algorithm
- features representation in XML config
- per-storage lookup for per-repository privileges
- dynamic SpringSecurity context update instead of delegating updated authentication through CustomAccessDecisionVoter


AlexOreshkevich:
Hi @carlspring, I would like to ask you starting reviewal process even if task it's not done. 

Please review my changes for security-users.xml, probably you have some better ideas how to organize new properties.

strongbox-user-management/src/main/java/org/carlspring/strongbox/users/domain/Features.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Map;
+
+import com.google.common.base.Objects;
+
+/**
+ * Additional user settings (for example, per-repository access settings).
+ *
+ * @author Alex Oreshkevich
+ */
+public class Features
```
carlspring: This is not a good name for the class. "Features" is very broad/generic and, at first glance, other developers won't be able to understand what this refers to.

strongbox-user-management/src/main/java/org/carlspring/strongbox/users/domain/User.java
```diff
@@ -50,6 +50,27 @@ public User(String id,
...
+               Objects.equal(salt, user.salt) &&
+               Objects.equal(roles, user.roles) &&
+               Objects.equal(securityTokenKey, user.securityTokenKey) &&
+               Objects.equal(features, user.features);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hashCode(username, password, enabled, salt, roles, securityTokenKey, features);
```
carlspring: Looks much neater, thanks!

strongbox-user-management/src/main/resources/etc/conf/security-users.xml
```diff
@@ -27,7 +27,62 @@
...
+        <username>deployer</username>
+        <credentials>
+            <password>password</password>
+            <encryption-algorithm>plain</encryption-algorithm>
+        </credentials>
+        <roles>
+            <role>UI_MANAGER</role>
+        </roles>
+
+        <features>
```
carlspring: I think this should be `<access-model>`, or something along these lines.

strongbox-user-management/src/main/resources/etc/conf/security-users.xml
```diff
@@ -27,7 +27,62 @@
...
+            <storages>
+                <storage id="storage0">
+                    <repositories>
+                        <repository id="act-releases-1">
```
carlspring: The most important part here is to be able to further control and fine-grain the access for the users based on paths in the repository. It should be possible to specify something like this here:

                                  <granted-paths>
                                       <path>com/carlspring/foo</path>
                                       <path>org/carlspring/foo</path>
                                  </granted-paths>

carlspring:
@AlexOreshkevich ,

So far, I think it looks promising, although I haven't had a chance to test it myself.

We need to be able to specify paths in the repository to which the user has been granted access. If no paths have been specified, then the user should have access to the whole repository.

One thing to note is that we'll need to have a controller around this, as it will have to somehow be configurable over the REST API.

strongbox-user-management/src/main/java/org/carlspring/strongbox/config/UsersConfig.java
```diff
@@ -157,6 +170,34 @@ private User toInternalUser(org.carlspring.strongbox.security.User user)
         internalUser.setRoles(user.getRoles());
         internalUser.setSalt(user.getSeed() + "");
 
+        // load features
+        org.carlspring.strongbox.security.Features features = user.getFeatures();
+        if (features != null)
```
sbespalov: I suggest to decrease the code complexity like this:
```
if (features == null){
    return internalUser;
}
```

sbespalov:
globally I suggest to use what Spring Security already have for this.
We can use `GrantedAuthority` with `authority=read:com/carlspring/foo` or `authority=write:com/carlspring/foo`

something this way instead of `Features` and other.

strongbox-user-management/src/main/java/org/carlspring/strongbox/users/domain/Features.java
AlexOreshkevich: Ok, sure. I will change it to UserAccessModel for now.

strongbox-user-management/src/main/java/org/carlspring/strongbox/config/UsersConfig.java
AlexOreshkevich: Disagree. Complexity will not be decreased in that case.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/CustomAccessDecisionVoter.java
```diff
@@ -0,0 +1,123 @@
...
+ * @author Alex Oreshkevich
+ * @see https://youtrack.carlspring.org/issue/SB-603
+ */
+public class CustomAccessDecisionVoter
+        implements AccessDecisionVoter, AuthenticationProvider
+{
+
+    private static final Logger logger = LoggerFactory.getLogger(CustomAccessDecisionVoter.class);
+
+    private Authentication authentication;
```
AlexOreshkevich: I knew that and I was looking into that way. The problem is that we have to deal with dynamic repositories and dynamic access rights. We can't hardcode URL's or any path's or anything because they are all subject of change.

Even if authority processing could be customized and custom rules and parsers could be registered I personally decided that my approach is simpler and more readable and extendable.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/CustomAccessDecisionVoter.java
```diff
@@ -0,0 +1,123 @@
...
+        if (features == null)
+        {
+            return vote;
+        }
+
+        // save request url during first authentication phase
+        // parse repositoryId from pre-saved during second authentication phase
+        if (object instanceof FilterInvocation)
+        {
+            user.setUrl(((FilterInvocation) object).getRequestUrl());
```
AlexOreshkevich: Kind of opening door in opposite direction but I agree here with you. It will look more readable.

strongbox-user-management/src/main/java/org/carlspring/strongbox/config/UsersConfig.java
sbespalov: Why do we need all these lines inside if? And what if there appears one more if inside it? 
So it will be complexity+1, and we can avoid it with above.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/CustomAccessDecisionVoter.java
sbespalov: > We can't hardcode URL's or any path's or anything because they are all subject of change.

i suppose that reposytory and storage urls will not change

carlspring:
@AlexOreshkevich,

* Is there any progress on this?
* The latest changes from the `master` will need to be picked up, as there was a massive refactoring in #227.

strongbox-user-management/src/main/resources/etc/conf/security-users.xml
```diff
@@ -27,7 +27,66 @@
             <encryption-algorithm>plain</encryption-algorithm>
```
carlspring: @AlexOreshkevich : Would you mind renaming this class to `strongbox-security-users.xml`?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
```diff
@@ -465,8 +463,7 @@ protected void doDeletePath(RepositoryPath repositoryPath,
     {
         Files.delete(repositoryPath);
 
-        Repository repository = repositoryPath.getFileSystem()
-                                              .getRepository();
```
sbespalov: @carlspring why you don't like such type of formatting? I think that it's more readable when invocation chain paced on new line.

strongbox-user-management/src/main/resources/etc/conf/security-users.xml
```diff
@@ -27,7 +27,66 @@
...
         </credentials>
         <roles>
-            <role>USER_ROLE</role>
+            <role>ADMIN</role>
         </roles>
-    </user>    
+    </user>
+    <user>
```
carlspring: @AlexOreshkevich ,

I know you knocked this up quickly for the sake of testing, but would it be possible to add a method to one of the base test classes, so that these can be configured programatically during tests? I mean -- we wouldn't like to be leaking test resources (which, in effect, this currently is) to the public. Clearly, there will be no `storage0:act-releases-1` outside of our tests. :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
carlspring: If it's a long list of these, sure, this should be chunked and aligned by the first method call and then the chained ones should all be on new lines. But... when it's a handful of short-named methods that fit on the same line, it looks odd to take 4-5 lines for something like:

    foo.bar()
       .do()
       .it();

instead of just:

    foo.bar().do().it();

If you guys feel strongly about it, then sure, we can adopt this. But I just don't want to see a lot of unnecessary code reformatting in pulls outside the blocks which are being worked on.

sbespalov: if you countning the line of codes then why do we need `{` on the next line? 
It's certainly just sarcasm :)
But do you mind if I will format code parts which I worked on like below? 
```
foo.bar()
   .do()
   .it();
```

carlspring: I'm not counting the lines of code at all... It's not about that. 

`for`/`do`/`while` statements should have curly braces. 

Classes and methods should have a curly brace on a new line.

I'm fine with

```
foo.bar()
   .do()
   .it();
```

What I'm not okay with is diff-s where the IDE, (or the developer), have done a "code reformat/re-ordering" on an entire class, when there are in fact ten lines of changes, and the operation has added another 50 changes. This is what I'm talking about.

strongbox-user-management/src/main/resources/etc/conf/security-users.xml
AlexOreshkevich: Done

PR #242, opened 2017-04-22
SB-786: Repository based `user-agent` mapping added to `HeadersMappingFilter`

sbespalov:


strongbox-web-core/src/main/java/org/carlspring/strongbox/web/HeaderMappingFilter.java
```diff
@@ -29,8 +31,13 @@
...
+import org.carlspring.strongbox.configuration.ConfigurationManager;
 import org.carlspring.strongbox.controllers.maven.MavenArtifactController;
 import org.carlspring.strongbox.controllers.nuget.NugetPackageController;
+import org.carlspring.strongbox.providers.layout.Maven2LayoutProvider;
+import org.carlspring.strongbox.providers.layout.NugetHierarchicalLayoutProvider;
+import org.carlspring.strongbox.storage.Storage;
+import org.carlspring.strongbox.storage.repository.Repository;
 
 import edu.emory.mathcs.backport.java.util.Arrays;
 import edu.emory.mathcs.backport.java.util.Collections;
```
carlspring: This looks like a set of incorrect imports...?

strongbox-web-core/src/main/java/org/carlspring/strongbox/web/HeaderMappingFilter.java
```diff
@@ -55,14 +62,27 @@
...
 
     @Override
     public void init(FilterConfig filterConfig)
         throws ServletException
     {
+        //Do nothing
+    }
+
+    @PostConstruct
+    public void postConstruct()
```
carlspring: It should be possible to populate this via some sort of registry, or something. Maybe the respective provider could have a `getUserAgent()`, or something... If the `LayoutProviderRegistry` is injected here, you could loop over the registered providers and get it a bit more dynamically, because at some point, hopefully, people might implement their own providers that are not part of the standard distribution and it will be hard to register them without such a dynamic approach.

carlspring:
Overall, the commit looks good! Could you please sort out the code style and the comments and I'll merge it when you're done...?

strongbox-web-core/src/main/java/org/carlspring/strongbox/web/HeaderMappingFilter.java
sbespalov: yes, it is
fixed

strongbox-web-core/src/main/java/org/carlspring/strongbox/web/HeaderMappingFilter.java
sbespalov: added TODO about this, I think this is not critical for now, and we can think how to implement such auto configuration. Maybe we don't need a `user-agent` header at all and should use some kind of custom header for this like `stromgbox-layout`

sbespalov:
@carlspring 
I'm very sorry, but could you specify places with unacceptable formatting?

PR #244, opened 2017-04-23
SB-787: When undeleting from Maven repositories, make sure the artifacts are re-added to the Maven Indexer

carlspring:


carlspring:
Thanks for your help, guys! :)

PR #253, opened 2017-05-12
SB-780: Add artifacts to the local Maven index when the repository is a proxied one

carlspring:


carlspring:
@fuss86,

Thanks for your help with this!

PR #271, opened 2017-05-24
SB-804 anonymous user problems fix

fuss86:


carlspring:
Thanks for you great help with this! :)

PR #276, opened 2017-05-31
SB-801 auth-class-loader problem fix

fuss86:


carlspring:
Thanks! :)

sbespalov:
@fuss86 I want to point out that the  Authenticator classes loading implementation done not very well now. Rather, even the current implementation is conceptually wrong in accordance with the fact that class loading in Java occurs dynamically at the time of accessing the class in the code. And this was the real root of problem you tried fix with this PR. Actually you don't fix the root, and it 100% will appear in the future, because you shouldn't load all and everything from any jar at once. 
To avoid the root issue and follow Java class loading concept you should use some `ClassPath` scanning tool to find Classes and Resources you need at first, and load only them after that. You can use something like [this](https://github.com/lukehutch/fast-classpath-scanner) for `ClassPath` scanning

carlspring:
I believe Sergey's got a point here. We should scan the matching jars, as this will make it quicker and less error-prone.

Also, @sbespalov, we made a small change that only picks up jars from the `webapp/WEB-INF/lib` that match a certain pattern.

sbespalov:
>Also, @sbespalov, we made a small change that only picks up jars from the webapp/WEB-INF/lib that match a certain pattern.

but it still loads everything from the jars at once. Generally you don't need this `jar` patterns at all, just scan what you need from all the class path and then load what was found.

fuss86:
@sbespalov, I see your point and can't disagree with you :) Loading classes is an overhead in our situation. We should let JVM loads necessary classes on demand, dynamically, as there is no need to do that explicitly.

I've created a PR for that: https://github.com/strongbox/strongbox/pull/278 

> you should use some ClassPath scanning tool to find Classes and Resources you need at first

I don't think it's necessary improvement right now. The issue here is we load classes unnecessarily, so I'm just leaving existing code that adds jar files to class loader. JVM will load them, on demand.

PR #279, opened 2017-06-02
SB-811: small NPE fix

sbespalov:
This fix is not directly related with SB-811 but it need to be done and can be merged.


carlspring:
Thanks, @sbespalov ! :)

PR #280, opened 2017-06-02
SB-811 responding HTTP 401 on access denied

fuss86:


strongbox-web-core/src/main/java/org/carlspring/strongbox/config/SecurityConfig.java
```diff
@@ -53,6 +55,8 @@ protected void configure(HttpSecurity http)
         http.sessionManagement()
             .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
             .and()
+            .exceptionHandling().authenticationEntryPoint(basicAuthenticationEntryPoint())
```
sbespalov: @fuss86 I think that `basicAuthenticationEntryPoint` shouldn't be hard coded like this, because it should be provided with `Authenticator` somehow. For example we can add the target `user-agent` for `Authenticator`, and choose `EntryPoint` this way. Also we can have a default `Authenticator` to use for `EntryPoint` and this can be a `BasicAuthAutenticator`.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/FilterAccessDecisionManager.java
```diff
@@ -1,30 +0,0 @@
-package org.carlspring.strongbox.security.vote;
```
sbespalov: why it was removed?

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/SecurityConfig.java
sbespalov: But I am not sure that `user-agent` is the best way, and maybe you will have some thoughts about this.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/FilterAccessDecisionManager.java
fuss86: I think yes, thanks to `MethodAccessDecisionManager` registered in SecurityConfig, but correct me if I'm wrong.

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/SecurityConfig.java
fuss86: Right now it fits your requirement to ask user for username/password as HttpBasic states. I think we can handle your improvements in the following tasks. 

@carlspring, what do you think ?

carlspring:
If it's functional enough at the moment, let's merge it and carry out the
follow-up work in a separate pull. This will help have a feeling of
accomplishment and make it look like things are progressing, which is
exactly what we have here.

If you're happy with this, I will merge it.

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/SecurityConfig.java
sbespalov: @fuss86 if you agree with sugestion then it would be great to add a `TODO` about this.

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/FilterAccessDecisionManager.java
sbespalov: I suppose that it related not only with metods, `SB-603` assumes URL access restriction too. This described in project Wiki [here](https://github.com/strongbox/strongbox/wiki/The-strongbox%E2%80%90authentication%E2%80%90users.xml-File#example-of-custom-user-access-model)

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/SecurityConfig.java
fuss86: @sbespalov, I've created new ticket SB-813 for that and placed `//TODO`

strongbox-web-core/src/main/java/org/carlspring/strongbox/security/vote/FilterAccessDecisionManager.java
fuss86: Generally, I did wrong that I removed this file in this branch since it's not strictly related with this task. I rolled it back. 

`FilterAccessDecisionManager `is not used currently, since @carlspring had some problems accessing repositories anonymously couple of days ago and I realized it was the main reason. I've created new ticket [SB-814](https://youtrack.carlspring.org/issue/SB-814) to determine its necessity.

From the first point of view `FilterAccessDecisionManager `is useless, since `CustomAccessDecisionVoter` (a thing responsible for what you pointed to:

> I suppose that it related not only with metods, SB-603 assumes URL access restriction too. 

) is also used by `MethodAccessDecisionManager` and URL access is checked there.

fuss86:
> If it's functional enough at the moment, let's merge it and carry out the follow-up work in a separate pull. This will help have a feeling of accomplishment and make it look like things are progressing, which is exactly what we have here. 

This PR contains only things that were required initially and meet requirements. Other things are written to: [SB-813](https://youtrack.carlspring.org/issue/SB-813) and [SB-814](https://youtrack.carlspring.org/issue/SB-814)

carlspring:
Thanks!

PR #288, opened 2017-06-14
make docs accessible

fuss86:


carlspring:
Thanks!

carlspring:
Could you please also close the task in Youtrack?

PR #313, opened 2017-07-20
SB-854-1 problem with closed stream

fuss86:


carlspring:
@sbespalov, it would be great if you could have a look just in case.  I will merge it now.

PR #325, opened 2017-08-06
SB-827: Extract all the layout provider-specific cron tasks to the respective layout provider implementation

carlspring:
- Committing work-in-progress.
- The unit tests pass.
- The integration tests for the cron tasks are broken. There are issues with loading the contexts more than once during testing.


carlspring:
Merged as part of #329 .

PR #327, opened 2017-08-10
releasing leased connections when we are done with the response

fuss86:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/ProxyRepositoryProvider.java
```diff
@@ -163,4 +169,23 @@ public ArtifactOutputStream getOutputStream(String storageId,
         return (ArtifactOutputStream) Files.newOutputStream(repositoryPath);
     }
 
+    private class CloseableProxyRepositoryResponse
```
fuss86: I don't want it to be static

carlspring:
Thanks a lot for fixing this! :)

PR #328, opened 2017-08-13
policy bugs fixes

fuss86:


carlspring:
Hi @fuss86 ,

It looks like you're on the right track, but the currently proposed fix is not how we should do it. It seems like we'll have to first implement a facility for managing version validators per repository. (You can have a look all the `*VersionValidator` classes).

At the moment, these are generally intended for Maven repositories. In general, we should support three policies -- `release`, `snapshot` and `mixed`. Obviously, `mixed` means the repository supports both `release` and `snapshot` artifacts. This would be useful for group repositories, because group repositories might consist of a mixture of repositories with either of the two (unless we have group repositories explicitly configured to work with just one of the two policies).

What I propose is to have a separate task (+pull request) that covers the implementation custom definition of these version validators in the `strongbox.xml`.

Let's discuss it over Skype.

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/maven/MavenArtifactControllerTest.java
```diff
@@ -841,15 +841,39 @@ public void testThatMavenProxyRepoDoesNotBlockTheConnectionPool()
...
                .then()
-               .statusCode(HttpStatus.NOT_FOUND.value());
+               .statusCode(HttpStatus.OK.value());
+    }
+
+    @Test
+    public void shouldDownloadProxiedSnapshotArtifact()
+            throws Exception
+    {
+        String url = getContextBaseUrl() +
```
carlspring: In time, this test will not be reproducible. The timestamps for snapshot artifacts graudally move on, and artifact repository managers remove older timestamps, which will at some point make this URL not work. For the time-being, this test will work, but at some point unexpectedly, it will start failing.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
```diff
@@ -1,14 +1,30 @@
...
+    SNAPSHOT("snapshot")
+            {
+                @Override
+                public boolean acceptsReleases()
+                {
+                    return false;
+                }
+            },
 
     MIXED("mixed");
```
carlspring: If you were to do things through this `enum`, then you would have to also implement these two methods for `MIXED` and return true for them, right...?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
```diff
@@ -25,9 +41,22 @@ public String getPolicy()
         return policy;
     }
 
-    public void setPolicy(String policy)
+    public boolean acceptsReleases()
     {
-        this.policy = policy;
+        return true;
```
carlspring: Defaulting to `true` here (and below) is not a very good idea, to be honest. Most repositories will have either a `snapshot`, or `release` policy. Usually, (statistically speaking), these are much more than the `group` repositories. Hence, defaulting to `false` would be more reasonable. Although, this, in a way, is cosmetic.

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/maven/MavenArtifactControllerTest.java
carlspring: Perhaps we could have some test artifacts deployed to a `snapshot` repository in strongbox that is included in a group through which the test case can resolve it directly against strongbox itself...? This should be more reproducible.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
fuss86: I'm not sure if I follow this comment. `true` is the default return value for RepositoryPolicyEnum `acceptsReleases() `method of constant values. Most of RepositoryPolicyEnum constant  values (actually 2 of 3) have this return value equal to `true`. That's why the default here is `true`. Would you like me to change it ?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
fuss86: No, there is a default value of `true` actually for these methods.

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/maven/MavenArtifactControllerTest.java
fuss86: you're right :)

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
carlspring: I'd recommend dropping the default `true` boolean values for the two fields below and having an explicit override of both methods. Makes it immediately clear when looking at the code, although it's still the same in effect.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
carlspring: As per my comment above, I think -- yes.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
fuss86: I'll replace it with a field

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
fuss86: I'll replace it with a field

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/maven/MavenArtifactControllerTest.java
fuss86: what do you propose @carlspring ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/validation/resource/ArtifactOperationsValidator.java
```diff
@@ -88,9 +88,9 @@ public void checkArtifactPath(String artifactPath)
     public void checkAllowsDeployment(Repository repository)
             throws ArtifactStorageException
     {
-        if (!repository.allowsDeployment() &&
-            !RepositoryTypeEnum.GROUP.getType().equals(repository.getType()) &&
-            !RepositoryTypeEnum.PROXY.getType().equals(repository.getType()))
+        if (!repository.allowsDeployment() ||
```
fuss86: Like this

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryPolicyEnum.java
```diff
@@ -1,33 +1,56 @@
...
+import java.util.stream.Stream;
+
 /**
  * @author mtodorov
  */
 public enum RepositoryPolicyEnum
 {
 
-    RELEASE("release"),
+    RELEASE("release", true, false),
```
carlspring: Cool, this looks a lot clearer! Thanks! :)

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/maven/MavenArtifactControllerTest.java
carlspring: Generate a snapshot Maven artifact in snapshot repository (see how it's done in other tests). Then include this repository in a group repository (if it's not already included) and just resolve it from there, instead of relying on an outside service. This way you'll always have the artifact.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -16,10 +16,10 @@
 public interface ArtifactManagementService extends ConfigurationService
 {
 
-    void store(String storageId,
-               String repositoryId,
-               String path,
-               InputStream is)
+    void validateAndStore(String storageId,
```
carlspring: I don't like this renaming very much. If we need to have two separate methods for this, then okay (one for validating and one for storing)... But actually, I think that the validation should be part of this method, so it should still be called `store`. Perhaps I'm not following your reasoning...? What are your thoughts?

fuss86: Honestly, I don't like it too.

Please compare [master version of AbstractArtifactManagementService.java](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/AbstractArtifactManagementService.java) with [updated in this branch AbstractArtifactManagementService.java](https://github.com/fuss86/strongbox/blob/ff35f4923aa89a593136e9265b0e3f75d7328060/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/AbstractArtifactManagementService.java) in terms of `store` and `validateAndStore` methods. 

In master branch, both `store` methods perform validation (`performRepositoryAcceptanceValidation`). Moreover, first `store` method invokes the second `store`, so the validation is performed twice unnecessarily. In master branch, the second `store` method is additionally invoked while downloading artifact from remote repository. As I explained it to you, validation (`performRepositoryAcceptanceValidation`) is also performed there in this case, over the remote repository, which is wrong. Those 2 cases show that the validation performed in the second `store` method is unnecessary and wrong. Therefore I removed it from this method in my branch.

But still first `store` method performs validation (`performRepositoryAcceptanceValidation`). So these 2 methods cannot be named the same. That's why I choose this name. 

What better solution do you suggest ?

carlspring: My apologies! I now see what you mean!
Okay, this is all good now! Thanks a lot! :)

carlspring:
@cs-devops-bot rebuild

PR #337, opened 2017-08-22
SB-852: Add an option to not download remote Maven indexes for proxy repositories

carlspring:


carlspring:
Thanks for your help, @fuss86 ! :)

PR #344, opened 2017-09-11
SB-615 configurable CORS

fuss86:


steve-todorov:
@carlspring I think you can now merge this to the master. Thanks for fixing this @fuss86! :)

PR #354, opened 2017-10-03
SB-887: remote  RSS index download for Nuget remote repository [in progress]

sbespalov:


strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/repository/RepositoryFeatures.java
```diff
@@ -11,5 +13,9 @@
     // Indexing implementation
 
     // List of supported checksums
-
+    
+    public void downloadRemoteIndex(String storageId,
```
carlspring: This should not be a method in the interface. In NuGet, this similar to a feed. In NPM and P2 there will be different things specific to those providers.

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/AbstractArtifactCoordinates.java
```diff
@@ -23,13 +23,19 @@ public AbstractArtifactCoordinates()
     public AbstractArtifactCoordinates(Map<String, String> coordinates)
     {
         this.coordinates = coordinates;
+        this.uuid = toPath();
```
carlspring: So... what's the purpose of this...?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/repository/NugetRepositoryFeatures.java
```diff
@@ -1,10 +1,112 @@
...
+            int packageCount;
+            try
+            {
+                packageCount = nugetClient.getPackageCount(false);
+            }
+            catch (IOException | URISyntaxException e)
+            {
+                throw new ArtifactTransportException(e);
+            }
+            for (int i = 0; i < packageCount / 100 + packageCount % 100; i++)
```
carlspring: Could you please explain (for my better understanding) what this loop is about?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/repository/RepositoryFeatures.java
sbespalov: I think that it should. 
> In NuGet, this similar to a feed. In NPM and P2 there will be different things specific to those providers.

all of then is indexes: 
- lucene for Maven
- OrientDB for Nuget

The interface is abstraction which is downloading "indexes" from some kind of "Remote source" and no matter what the indexes are and from what source.

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/AbstractArtifactCoordinates.java
sbespalov: the purpose that `Path` is unique and can be used as `uuid`

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/repository/NugetRepositoryFeatures.java
sbespalov: it's the iteration for Feed pages

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/repository/RepositoryFeatures.java
carlspring: Yes, these are indexes in the case of Lucene and OrientDB. However, for NuGet, they're feeds. I'm not sure what they're called in NPM-speak...

This is why I have implemented the concept of repository features.

We should not assume they're interchangeable terms and that each layout provider will have an index. In this particular case, we're not talking about OrientDB, or the Maven Indexer. These are, in effect different concepts.

sbespalov: In this case, I'm not sure that I fully understand for now this Repository Features concept, but I will remove method from interface as you wished.

sbespalov: Should I leave the method in `NugetRepositoryFeatures` class or move it somewhere else?

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/AbstractArtifactCoordinates.java
carlspring: The path is unique to the repository. I'm not sure, if this is relevant... (Just thought of mentioning it, just in case).

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/test/java/org/carlspring/strongbox/storage/indexing/NugetRemoteRepositoryTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Sergey Bespalov
+ *
+ */
+@ContextConfiguration(classes = NugetLayoutProviderTestConfig.class)
+@RunWith(SpringJUnit4ClassRunner.class)
+public class NugetRemoteRepositoryTest
+        extends TestCaseWithRepository
+{
+
+    private static final String REPOSITORY_RELEASES = "m2pr-releases";
```
carlspring: The convention that I've used is that each test should have their own repository. The repository's name should be an abbreviation of the class name's initials. In this case -- `nrrt`.

We need to keep making sure that each of the tests is omnipotent and that none of the test depend on each other.

The same applies for the variable below.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/repository/NugetRepositoryFeatures.java
```diff
@@ -1,10 +1,112 @@
...
         implements RepositoryFeatures
 {
 
+    @Inject
+    private ConfigurationManager configurationManager;
+
+    @Inject
+    private ArtifactEntryService artifactEntryService;
+
+    public void downloadRemoteIndex(String storageId,
```
carlspring: In this case, I assume a better name for this method, would probably be `downloadRemoteFeed`...? This is what I was referring to in my earlier comment regarding the method that you defined in the interface -- in one implementation, the "features" may be called `X`, `Y`, `Z`, while in another -- there could be `Y` and `A`. The `ReposiotryFeatures` interface defines no methods for this exact reason. However, there is a mechanism in place that uses these "features".

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/AbstractArtifactCoordinates.java
sbespalov: yes, you are right.
seems that i missed that

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/repository/NugetRepositoryFeatures.java
sbespalov: ok, I will rename it
thanks

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
```diff
@@ -18,7 +21,10 @@
     private String repositoryId;
 
     // if you have to rename this field please update ArtifactEntryServiceImpl.findByCoordinates() implementation
-    private ArtifactCoordinates artifactCoordinates;
+    @OneToOne(orphanRemoval = true)
+    private AbstractArtifactCoordinates artifactCoordinates;
+    
+    private String artifactPath;
```
sbespalov: @carlspring you may want to note this denoramalization. This is needed to create unique index for ArtifactEntry.  The index in turn is needed in order to quickly determine the presence of the required entity.

sbespalov: I was not able to find the best way :(

strongbox-commons/src/main/java/org/carlspring/strongbox/artifact/coordinates/AbstractArtifactCoordinates.java
```diff
@@ -5,6 +5,8 @@
 
 import org.carlspring.strongbox.data.domain.GenericEntity;
 
+import edu.emory.mathcs.backport.java.util.Collections;
```
carlspring: This is a incorrect import.

carlspring:
Thanks! :)

PR #355, opened 2017-10-04
SB-887: UnitTests issues fixed

sbespalov:


steve-todorov:
@carlspring I've tested this and it seems to be working fine now. You can merge it into the master. Thanks for fixing this @sbespalov! :)

carlspring:
Thanks for this fix!

PR #356, opened 2017-10-04
SB-562 preemptive maven-metadata.xml downloading at the artifact level

fuss86:


strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.maven.index.locator.MetadataLocator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+@Component
+public class MavenArtifactDownloadedEventListener
```
carlspring: This looks more, or less along the lines of what I had in mind, thanks! :)

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventListenerRegistry.java
```diff
@@ -244,4 +275,64 @@ public void dispatchArtifactArchivedEvent(String storageId,
...
+        logger.debug("Dispatching ArtifactEventTypeEnum.EVENT_ARTIFACT_METADATA_DOWNLOADING event for " +
+                     storageId + ":" + repositoryId + "/" + path + "...");
+
+        dispatchEvent(event);
+    }
+
+    public void dispatchArtifactChecksumDownloadedEvent(String storageId,
+                                                         String repositoryId,
+                                                         String path)
+    {
```
carlspring: Could you please just re-indent the parameters? Thanks.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventListenerRegistry.java
```diff
@@ -244,4 +275,64 @@ public void dispatchArtifactArchivedEvent(String storageId,
...
+        logger.debug("Dispatching ArtifactEventTypeEnum.EVENT_ARTIFACT_METADATA_DOWNLOADED event for " +
+                     storageId + ":" + repositoryId + "/" + path + "...");
+
+        dispatchEvent(event);
+    }
+
+    public void dispatchArtifactMetadataDownloadingEvent(String storageId,
+                                                        String repositoryId,
+                                                        String path)
+    {
```
carlspring: Could you please just re-indent the parameters? Thanks.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventListenerRegistry.java
```diff
@@ -244,4 +275,64 @@ public void dispatchArtifactArchivedEvent(String storageId,
         dispatchEvent(event);
     }
 
+    public void dispatchArtifactMetadataDownloadedEvent(String storageId,
+                                              String repositoryId,
+                                              String path)
+    {
```
carlspring: Could you please just re-indent the parameters? Thanks.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
```diff
@@ -0,0 +1,147 @@
...
+import java.io.InputStream;
+import java.nio.file.NoSuchFileException;
+import java.security.NoSuchAlgorithmException;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.IOUtils;
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.repository.metadata.Metadata;
+import org.apache.maven.index.locator.Locator;
+import org.apache.maven.index.locator.MetadataLocator;
```
carlspring: I'm not sure why you're using:
```
import org.apache.maven.artifact.repository.metadata.Metadata;
import org.apache.maven.index.locator.Locator;
```

These are used for rebuilding indexes, as far I'm aware.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
```diff
@@ -0,0 +1,147 @@
...
+    public void handle(ArtifactEvent event)
+    {
+        if (event.getType() != ArtifactEventTypeEnum.EVENT_ARTIFACT_FILE_DOWNLOADED.getType())
+        {
+            return;
+        }
+
+        Repository repository = getRepository(event);
+
+        if (!RepositoryLayoutEnum.MAVEN_2.getLayout().equals(repository.getLayout()))
```
carlspring: This should actually probably be the first check of the method...

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
```diff
@@ -0,0 +1,147 @@
...
+            return;
+        }
+
+        resolveArtifactMetadataFile(event);
+    }
+
+
+    private void resolveArtifactMetadataFile(ArtifactEvent event)
+    {
+        try
```
carlspring: I'm sorry, but I'm not quite sure I'm following your code very well...

What's going on over here? What are you trying to locate? Is this a check to see, if the artifact is on the file system?

This should be done via the `LayoutProvider` abstraction, because it handles the file system in a generic way (making it easier to implement Amazon S3 and Google Cloud Storage support later).

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
fuss86: > import org.apache.maven.artifact.repository.metadata.Metadata;

is for rebuilding indexes ?

> import org.apache.maven.index.locator.Locator;
> import org.apache.maven.index.locator.MetadataLocator;

In these cases, you're right, I shouldn't have used them. I haven't noticed that. I'll use more proper tools for my requirements.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
fuss86: No problem, whatever you like.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
fuss86: What does `resolveArtifactMetadataFile` method do:
1. find out if there is an existing `maven-metadata.xml` file for downloaded artifact
2. if there is no such file in the file system, download `maven-metadata.xml` file from the remote repository and store it in the file system, following SB-562 requirements
3. if there is such file in the local file system, download `maven-metadata.xml` and merge it with the existing  `maven-metadata.xml` file in the file system

> This should be done via the LayoutProvider abstraction
if you mean checking  if the artifact is on the file system, OK, I can implement it using `LayoutProvider`. Please confirm.

fuss86: On the other hand, this is strange a little bit. Why do you want to use `LayoutProvider` abstraction if we want it to be working in the `Maven` case ? `maven-metadata.xml` is strict `Maven` thing, generally we are trying to locate `maven-metadata.xml` file, not an artifact in general. Can't we use maven-related tools to locate  `maven-metadata.xml` file ?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
```diff
@@ -0,0 +1,147 @@
...
+        {
+            final File metadataFile = metadataLocator.locate(new File(event.getPath()));
```
sbespalov: Sorry guys, but can we use `Path`  instead of `File` everywhere?
You can use `layoutProvider.resolve(repository).resolve(artifactLocation)`to get `RepositoryPath` you need within `Repository`.

sbespalov: > This should be done via the LayoutProvider abstraction, because it handles the file system in a generic way (making it easier to implement Amazon S3 and Google Cloud Storage support later).

I mean @carlspring means my above comment here.

carlspring: Yes, this is correct. It seems I have failed to explain what I meant. We should be using `Path` instead of `File` in order to be ready to implement a cloud-based storage provider in the future (we currently only really have `FileSystemStorageProvider`).

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
carlspring: > import org.apache.maven.artifact.repository.metadata.Metadata;
> 
> > is for rebuilding indexes?

No, this is fine. I was referring to the `import org.apache.maven.index.locator.*` imports.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
carlspring: Thanks for your clear explanation!

carlspring: ...And now, regarding your questions -- as explained by @sbespalov , we should try to use `Path` as much as possible. I am not 100% sure, if the Maven Metadata classes that you're referring to can work with `Path`-s. If they can, then very well -- go ahead and use them, but I suspect they actually don't.

On the other hand, in the project we have `AbstractMavenArtifactLocatorOperation`, which I have feeling (off the top of my head) will cover what you need.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
fuss86: `MetadataHelper` contains many methods working on `java.io.File` making this helper class partially useless, right ? :(

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactDownloadedEventListener.java
fuss86: To resolve the metadata file on the artifactId level I'll have to use below construction:

```java
Repository repository = getRepository(event);
LayoutProvider layoutProvider = layoutProviderRegistry.getProvider(repository.getLayout());
RepositoryPath artifactPath = layoutProvider.resolve(repository).resolve(event.getPath());
Path metadataPath = artifactPath.getParent().getParent().resolve("maven-metadata.xml");
```

carlspring:
Thanks! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
```diff
@@ -0,0 +1,152 @@
...
+            InputStream is = response.readEntity(InputStream.class);
+            if (is == null)
+            {
+                return null;
+            }
+
+            is = onSuccessfulProxyRepositoryResponse(is, storageId, repositoryId, path);
+
+            final RepositoryPath artifactPath = layoutProvider.resolve(repository).resolve(path);
+            if (!layoutProvider.isChecksum(artifactPath) && !layoutProvider.isMetadata(path))
```
sbespalov: @fuss86 I think that it's better to use `RepositoryFileAttributes` here like follows:
```
    RepositoryFileAttributes a = (RepositoryFileAttributes) Files.readAttributes(path, BasicFileAttributes.class);
    if (!a.isChecksum() && !a.isMetadata())
    {
        ...
    }
```

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventListenerRegistry.java
```diff
@@ -110,7 +152,9 @@ public void dispatchArtifactDownloadingEvent(String storageId, String repository
         dispatchEvent(event);
     }
 
-    public void dispatchArtifactDownloadedEvent(String storageId, String repositoryId, String path)
+    public void dispatchArtifactDownloadedEvent(String storageId,
```
sbespalov: Am I correctly understood that he would be called twice?
once in `MavenArtifactController` and another in `ProxyRepositoryArtifactResolver`
Maybe it will be better to use different event type for Proxy Repositories?

sbespalov: @carlspring  and why do we have this events only in `MavenArtifactController`, they will not work for Nuget in this case.

fuss86: > Am I correctly understood that he would be called twice?
once in MavenArtifactController and another in ProxyRepositoryArtifactResolver
Maybe it will be better to use different event type for Proxy Repositories?

It looks so, I haven't noticed it. @carlspring , when should we dispatch artifact downloaded event ? When the artifact is downloaded from the proxy repository ? Or, when the artifact has been requested and flushed from the Controller ? 

Thanks, @sbespalov

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
fuss86: @carlspring, should I proceed with this change in some another pull request ?

carlspring: Yes, please.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventListenerRegistry.java
carlspring: Hi,

Thanks for spotting this, @sbespalov!

@fuss86: I believe the better option would be #2 -- after it's flushed. What do you guys think?

fuss86: Ok, I'll add one new event when the proxied file has been downloaded from remote repository :)

PR #359, opened 2017-10-12
fixing problems with indexing access model

fuss86:


strongbox-security/strongbox-user-management/src/main/java/org/carlspring/strongbox/users/domain/User.java
```diff
@@ -123,6 +124,14 @@ public AccessModel getAccessModel()
 
     public void setAccessModel(AccessModel accessModel)
     {
+        // ORecordDuplicatedException: Cannot index record AccessModel{....}:
+        // found duplicated key '....' in index 'idx_uuid' previously assigned to the record ...
+        if (this.accessModel != null && accessModel != null && accessModel.getUuid() != null &&
+            Objects.equal(this.accessModel.getUuid(), accessModel.getUuid()) &&
+            !Objects.equal(this.accessModel.getObjectId(), accessModel.getObjectId()))
+        {
+            accessModel.setUuid(UUID.randomUUID().toString());
```
sbespalov: I am not sure that this logic will be proper.
i think that it should be like follows:
```
if (this.accessModel != null)
{
    acessModel.setUuid(this.accessModel.getUuid());
    acessModel.setObjectId(this.accessModel.getObjectId());
}
```

PR #360, opened 2017-10-15
record ID should be set cascade entities for known uuid entities

fuss86:


sbespalov:
@fuss86  thanks for this
@carlspring it looks good to me too, and can be merged

carlspring:
@fuss86 : Thanks for your fixes!
@sbespalov : Thanks for reviewing!

PR #384, opened 2017-11-20
SB-930 orientDb cascade updates does not work [idle TODO]

fuss86:


sbespalov:
implemented in https://github.com/strongbox/strongbox/pull/620

PR #385, opened 2017-11-20
SB-932 xml configurable cron jobs

fuss86:


strongbox-configuration/src/main/java/org/carlspring/strongbox/xml/PropertiesAdapter.java
```diff
@@ -0,0 +1,55 @@
+package org.carlspring.strongbox.xml;
+
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import com.sun.org.apache.xerces.internal.dom.DocumentImpl;
```
carlspring: We shouldn't be using any com.sun.* imports.

strongbox-cron/strongbox-cron-api/src/main/java/org/carlspring/strongbox/cron/domain/CronTaskConfiguration.java
```diff
@@ -105,4 +110,15 @@ public void setImmediateExecution(boolean immediateExecution)
         this.immediateExecution = immediateExecution;
     }
 
+    @Override
+    public String toString()
+    {
+        return new ToStringBuilder(this)
```
carlspring: I've actually really been thinking about this a lot lately. Everybody seems to be doing their own thing with the `toString()` methods. Perhaps we need to vote on a library that can generate these for us, do a refactoring and start being more consistent?

strongbox-cron/strongbox-cron-rest/src/test/resources/etc/conf/strongbox-cron-tasks.xml
```diff
@@ -0,0 +1,117 @@
...
+            <jobClass>org.carlspring.strongbox.cron.jobs.RegenerateChecksumCronJob</jobClass> <!-- mandatory -->
+            <storageId>storage-common-proxies</storageId> <!-- optional -->
+            <repositoryId>maven-central</repositoryId> <!-- optional -->
+            <basePath>org</basePath> <!-- optional -->
+            <forceRegeneration>true</forceRegeneration> <!-- optional, default false -->
+        </properties>
+    </cron-task-configuration>
+
+    <cron-task-configuration>
+        <name>DownloadRemoteMavenIndexCronJob</name>
```
carlspring: I think we need some sort of format for these... Especially, if they're system generated ones. Perhaps something like "Download remote Maven Indexes for ${storageId}:${repositoryId}".

I mean... this XML file will be the one going into production. :)

carlspring:
This pull looks great, thanks! :)

I have some minor remarks in the code. Also, there seems to be a failing test case. Would you mind addressing these when you can?

carlspring:
Thanks for your fixes! :)

PR #394, opened 2017-11-28
SB-624: When a remote is not reachable, do not expire the cache

fuss86:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/LocalStorageProxyRepositoryExpiredArtifactsCleaner.java
```diff
@@ -52,18 +58,52 @@ public void cleanup(final Integer lastAccessedTimeInDays,
...
+            return;
+        }
+        for (final Iterator<ArtifactEntry> it = artifactEntries.iterator(); it.hasNext(); )
+        {
+            final ArtifactEntry artifactEntry = it.next();
+            final Storage storage = configurationManager.getConfiguration().getStorage(artifactEntry.getStorageId());
+            final Repository repository = storage.getRepository(artifactEntry.getRepositoryId());
+            if (!repository.isProxyRepository())
+            {
+                logger.debug("Repository {} is not a proxy repository.", repository.getId());
```
carlspring: I think will be causing a lot of unnecessary debugging output and we could probably safely remove it.

PR #443, opened 2017-12-23
SB-962: LoginController fixed

sbespalov:


carlspring:
Thanks a lot for your quick fix!

Issue #408, opened 2017-12-09
Upgrade org.apache.httpcomponents:* to >= 4.5.4


carlspring:
We need to upgrade to the latest version of Jetty.

Make sure that:
* The project still builds
* The tests are still passing

The version is controlled via the `strongbox-parent/pom.xml`.

dinesh19aug:
Created a pull request with #366 fix. Updated the version to 4.5.4. The tests were successful.

dinesh19aug:
Rolled back changes. The change builds fine(all test cases pass) on my machine but fails in Jenkins

carlspring:
BTW, you may want to join our [chat](https://gitter.im/strongbox/strongbox).

carlspring:
@dinesh19aug,

BTW, you may want to join our [chat](https://gitter.im/strongbox/strongbox).

carlspring:
Blocked by https://github.com/jersey/jersey/issues/3629.

carlspring:
There is now a `4.5.5`.

carlspring:
@ptirador,

Would you like maybe to look into this?

ptirador:
Yeah, sure! I'll look into it.

carlspring:
Fixed by @ptirador! :)

PR #451, opened 2017-12-25
Upgrade latest version http-components

epidemia:
This pull is for issue #408.


epidemia:
There is a problem in jersey when using http-components:4.5.+.
https://github.com/jersey/jersey/issues/3629

carlspring:
Well, in this case we'll have to wait for this to be fixed before we could upgrade.

PR #452, opened 2017-12-26
SB-963: CORS fixed in case of JwtTokenValidationFilter [ready]

sbespalov:


steve-todorov:
@carlspring this works as expected and can be merged. Thanks @sbespalov! :)

carlspring:
Good job! :)
Thanks!

PR #457, opened 2017-12-27
SB-974 WWW-Authenticate not returned when X-Requested-With: XMLHttpRe…

fuss86:
…quest is present in the request


steve-todorov:
This is working great! Thanks @fuss86 for fixing this! 

@carlspring proceed with merging :)

PR #468, opened 2018-01-02
Issue 370 2

dinesh19aug:


strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
```diff
@@ -40,17 +49,31 @@
...
         {
             addAlgorithm(algorithm);
         }
+
+    }
+
+    private boolean isInputStreamValid()
+            throws IOException
+    {
+        if (in.available()>0)
```
carlspring: Please, always use braces, even for one-liners. Could you please change this across your pull?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
```diff
@@ -152,6 +177,10 @@ public int read(byte[] bytes,
                 MessageDigest digest = (MessageDigest) entry.getValue();
                 digest.update(bytes, off, numberOfBytesRead);
             }
+            if(isInputStreamValid()){
```
carlspring: Cosmetic: We use braces on new lines. Could you please change this?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
```diff
@@ -152,6 +177,10 @@ public int read(byte[] bytes,
...
+                this.extension = getFileExtension(bytes);
```
carlspring: The problem with this approach is that you have to have read the whole file in memory in order to pass the array of bytes. How much memory would you need to have, if you had 50 different sessions deploying artifacts each of around 200-300 MB (let's assume they're `war` files). This would be very memory inefficient, won't it?

Couldn't we just have some sort of `InputStream` wrapper, or something, that just uses the bytes that have already been ready (which we're already passing to the `digest.update((byte)...)` method? If we could implement something like this, then things would be significantly more efficient and lightweight.

What do you think?

dinesh19aug:
Reformatted the code. 
The way the code works is that whenever the method byte[] bytes, int off, int len) is called during the process of when file is passed as input stream and read, the getExtension method is called. There is a utility method which is checked before trying to find the extension -- "isInputStreamValid". The method basically looks ahead to see if it is not an empty file. The reason this method is there to check file validity is that some of the test cases are testing negative scenarios by passing non existent file or invalid file, if this method is not in place, the code will throw error that file cannot be found or invalid file.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
dinesh19aug: Fixed.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
dinesh19aug: Fixed

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/io/ArtifactInputStream.java
dinesh19aug: That's what I had been trying to for last 10 days, but the problem is that no matter what warpper you choose BufferedOutputStream/inputStream, the moment you try and read those bytes, inputStream becomes invalid for some of the web-core test like "testMetaDataMerge" or "testPackageCommonFlow" as the stream may not support mark and reset.

As far as trying to avoid memory buffer goes, whether we create a wrapper object like ByteOutputStream --> convert to bytes --> BufferedInoutStream for example, will still take up memory space when you are trying to convert to a parallel stream.

I have used this code before to read the stream twice but here in this project, I get message that some of the test based on some test resources, the files are not mark() supoorted.
`ByteArrayOutputStream baos = new ByteArrayOutputStream();
org.apache.commons.io.IOUtils.copy(in, baos);
byte[] bytes = baos.toByteArray();

// either
while (needToReadAgain) {
    ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
    yourReadMethodHere(bais);
}

// or
ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
while (needToReadAgain) {
    bais.reset();
    yourReadMethodHere(bais);
}`

So if we want to use a parallel stream then we will need to revisit the test cases in web-core domain. The sample test resource  files that are being used there have issues and do not support mark and reset.

strongbox-commons/src/test/resources/pom.properties
```diff
@@ -0,0 +1,4 @@
+#Created by Apache Maven 3.3.9
```
carlspring: Generally, such test files should be produced by your tests, in order to make sure that things are reproducible and that each test has the resources it needs. Otherwise, when you look at the resources themselves, you have no clue which test owns them. This is a good testing practice to follow.

strongbox-commons/src/test/resources/testTextFile.txt
```diff
@@ -0,0 +1 @@
+This is a text file.
```
carlspring: Same as my comment above, regarding test resources.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/resources/artifacts/pom.properties
```diff
@@ -0,0 +1,4 @@
+#Created by Apache Maven 3.3.9
```
carlspring: Same as my earlier comment regarding test resources.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/resources/artifacts/testTextFile.txt
```diff
@@ -0,0 +1 @@
+This is a text file.
```
carlspring: Same as my earlier comment regarding test resources.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/io/ArtifactInputStreamTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void testFileExtensionEqualsGz()
+            throws FileNotFoundException, NoSuchAlgorithmException
+    {
+        try
+        {
+            inputStream = new BufferedInputStream(new FileInputStream(pathSh.trim()));
+            artifactInputStream = new ArtifactInputStream(null, inputStream)
+            {
+            };
+            assertTrue(artifactInputStream.getFileExtension(IOUtils.toByteArray(inputStream)).equalsIgnoreCase(".gz"));
```
carlspring: This is incorrect. The extension here should actually be `tar.gz` (and with no dot).

dinesh19aug: hmm.. I will check if Tika has another method. Currently the detect method returns .gz as file type.

carlspring:
Asked for advice on [Stackoverflow](https://stackoverflow.com/questions/48102004/how-to-implement-an-inputstream-that-dynamically-guesses-the-extension-of-a-file) and the Apache Tika User list.

Issue #464, opened 2017-12-29
Make it possible to list storages and their repositories when browsing


carlspring:
At the moment, it's not possible to list storages, or repositories, just the content of repositories.

It would be great to have this implemented.

This relates to #418, #419, #420, #444.

sanket407:
Working on this !!

carlspring:
Thanks for implementing this! :)

PR #469, opened 2018-01-02
#464 - Added browsing support for Storages and Repositories under a Storage

sanket407:
Implemented #464 
1. Added REST endpoints for browsing storages and repostories in BrowseStoragesController in org.carlspring.strongbox.controllers  

- /storages  - Browse all the storages 
- /storages/STORAGE - Browse all the repositories under Storage - 'STORAGE' 

For directly Invoking or Testing Against this controller set header 'user-agent=unknown/*'

2. Added code for redirecting a GET request for a directory not ending in '/' to the equivalent path ending in '/'. So /storages will get redirected to /storages/ , /storages/storage0' will get redirected to /storages/storage0/' .


steve-todorov:
I've tested this and it is working fine. @carlspring any thoughts on the logging (should we drop something or leave it as is) ? 

Good job @sanket407! :)

carlspring:
Thanks! :)

PR #479, opened 2018-01-06
Issue 462

dinesh19aug:


strongbox-metadata-maven-api/src/main/java/org/carlspring/strongbox/storage/metadata/MetadataMerger.java
```diff
@@ -108,19 +128,151 @@ public Metadata updateMetadataAtGroupLevel(PluginArtifact artifact, Metadata met
...
+    /**
+     * This method will extract plugin.xml from artifact archive and save it in the temp directory
+     *
+     * @param zipEntry
+     * @param artifact
+     * @param zipInputStream
+     * @return
+     * @throws IOException
+     */
+    private boolean extractPluginXmlFile(ZipEntry zipEntry,
```
carlspring: You should not be extracting this file. You can just grab it's `InputStream` from the zip.

dinesh19aug: I thought of doing the same and keeping it in BufferedOutputStream but the problem with that approach is that if too many streams are running are running in parallel, it will eat up memory.

carlspring:
Asked "[how to parse META-INF/maven/plugin.xml](https://stackoverflow.com/questions/48125165/how-to-parse-meta-inf-maven-plugin-xml-files-using-the-maven-api-s)" files on Stackoverflow.

carlspring:
Could you please replace the SAX parsing with the [PluginMetadataParser](https://github.com/apache/maven-plugin-tools/blob/a7fc242de5edca778f521150a80724bb586c4851/maven-script/maven-plugin-tools-model/src/main/java/org/apache/maven/tools/plugin/extractor/model/PluginMetadataParser.java)?

strongbox-metadata-maven-api/src/main/java/org/carlspring/strongbox/storage/metadata/MetadataMerger.java
```diff
@@ -108,19 +128,96 @@ public Metadata updateMetadataAtGroupLevel(PluginArtifact artifact, Metadata met
...
+        HashMap<String, String> pluginMap = null;
+        try
+        {
+            fileInputStream = new FileInputStream(artifact.getFile().getCanonicalPath());
+            zipInputStream = new ZipInputStream(fileInputStream);
+
+
+            while ((zipEntry = zipInputStream.getNextEntry()) != null)
+            {
+                if (zipEntry.getName().endsWith("plugin.xml"))
```
carlspring: I think there should be a more explicit check for the exact path to the `META-INF/maven/plungin.xml`, because... what will happen, if, say, there is coincidentally some other `plugin.xml` file...?

dinesh19aug: Replaced the check for plugin.xml with a more comprehensive check for plugin.xml path under META-INF/maven/plugin.xml

dinesh19aug:
Is there anything else pending on this? PluginMetadataParser as already informed is not the right class and throws error that it cannot parse. The ticket said that Zipinuputxxx should help us get the plugin.xml and later we need to parse the file to get the values. The current code uses SAX to do the same.

carlspring:
Hi,

Sorry, I've had a few pull requests on the list.
The code looks good. I investigated Maven's API and you are indeed right that their parser is not actually devised to read the XML as it is, it actually just populates certain fields of the object, which is quite incredible, to say the least...

I will merge this! Thanks for your help and patience!

Martin

Issue #477, opened 2018-01-05
Make the EnvironmentInfoController controller also display JVM args used to run Strongbox


carlspring:
In continuation of #458, make the `EnvironmentInfoController` also show the JVM args used to start Strongbox.

carlspring:
Implemented by @ptirador.

PR #480, opened 2018-01-06
Make the EnvironmentInfoController controller also display JVM args used to run Strongbox

ptirador:
Related issue: #477 

The /configuration/environment/info endpoint allows to fetch JVM args (among others) in JSON format, sorted.

Considered that a JVM argument is defined as -Dname=value or -Dname (in this case, empty string value).

I also implemented the proper test cases.


ptirador:
Updated to change the JSON format for JVM args.

`"jvm": [
     "-Dsome.property=true",
     "-Psome-profile",
     "-Xms1gb",
     ...
  ]`

The sorting has also changed: now it is case insensitive.

steve-todorov:
@ptirador thanks for your work! I've tested this and it's showing the arguments.  :) 
@carlspring I think we can merge this if you don't have anything more to add.

carlspring:
Thanks, guys!

Issue #445, opened 2017-12-24
Refactor UserController to return proper JSON messages upon successful/failed action


steve-todorov:
Refactor `org.carlspring.strongbox.controllers.users.UserController` (located in `strongbox-web-core`) so that it returns `JSON` messages upon successful/failed actions with the appropriate messages as follows:

1. If an action is successful we need to return the following JSON:
   ```
   {
       "message": "User successfully updated / deleted / etc"
   }
   ```

2. If the action has failed we need to return an error message that describes what's wrong:
   ```
   {
       "message": "A user with this username already exists! Please enter another username. / etc"
   }
   ```

carlspring:
Implemented by @ptirador.

PR #481, opened 2018-01-09
SB-445: Refactor UserController to return proper JSON messages upon successful/failed action

ptirador:
Related issue: #445 

In the method `create`, an additional check has been added before saving: if you try to create a user with an existing username, a 409 code (conflict) is returned with the proper body message wrapped.

I also implemented the proper test cases.


steve-todorov:
@ptirador thanks for your work! I've tested this and it works great.

@carlspring you can merge now :)

carlspring:
Thanks for your contribution!

Issue #372, opened 2017-10-25
Implement HEAD requests support for artifacts


carlspring:
We need to be able to respond to `HTTP HEAD` requests issued against the `org.carlspring.strongbox.controllers.maven.MavenArtifactController` and `org.carlspring.strongbox.controllers.nuget.NugetPackageController` classes.

Possible headers to return:
- `Content-Length`
- `Checksum-MD5`
- `Checksum-SHA1`
- `Creation-Date`
- `Last-Updated`

sanket407:
Working on this .

carlspring:
@sanket407 ,

As per your request, I have updated the task's description.

PR #491, opened 2018-01-13
Added header support for Maven and Nuget - Hosted Repositories #372 [ready]

sanket407:
#372 
Added Header Support for Maven and Nuget Hosted Repositories.
More code can be removed from controller after implementing getPath() for group and proxy repository.
Currently keeping it simple for hosted repository support.
Also GET doesn't return Last-Modified as of now as implementing it will either cause unnecessary code or GET failures for proxy and group repositories. Will implement it after getPath() is implemented for both group and proxy repositories . 
Currently GET and HEAD request for HostedRepositoryProvider will return same headers except Last-Modified.

### DIFF
--- a/strongbox-rest-client/src/test/java/org/carlspring/strongbox/rest/client/RestAssuredArtifactClient.java
+++ b/strongbox-rest-client/src/test/java/org/carlspring/strongbox/rest/client/RestAssuredArtifactClient.java
@@ -366,7 +366,7 @@ public class RestAssuredArtifactClient
                            .asString();
     }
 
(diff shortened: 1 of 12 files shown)


carlspring:
@sanket407 ,

Does this pull include the fixes for all the things raised by @fuss86 and @sbespalov in the previous pulls that you created?

sanket407:
@carlspring  ok  I will do it for npm too. Push it here itself.

carlspring:
Thanks, @sanket407!

Issue #509, opened 2018-01-18
Update the licensing information


carlspring:
* All the `LICENSE` files should have the year changed to 2018:
```
Copyright 2018, Carlspring Consulting & Development Ltd.
```
* All the `pom.xml` files should have the `<inceptionYear>` set to 2018.

The same changes should be applied to all of the projects in the [Strongbox](https://github.com/strongbox) organization on Github:
* ~~[strongbox-assembly](https://github.com/strongbox/strongbox-assembly)~~ _[done in [strongbox-assembly #9](https://github.com/strongbox/strongbox-assembly/pull/9)]_
* ~~[strongbox-webapp](https://github.com/strongbox/strongbox-webapp)~~ _[done in [strongbox-webapp #13](https://github.com/strongbox/strongbox-webapp/pull/13)]_
* ~~[strongbox-web-integration-tests](https://github.com/strongbox/strongbox-web-integration-tests)~~ _[done in [strongbox-web-integration-tests #19](https://github.com/strongbox/strongbox-web-integration-tests/pull/19)]_
* ~~[strongbox-examples](https://github.com/strongbox/strongbox-examples)~~ _[done in [strongbox-examples #9](https://github.com/strongbox/strongbox-examples/pull/9)]_  
* ~~[strongbox-authentication-example](https://github.com/strongbox/strongbox-authentication-example)~~ _[done in [strongbox-authentication-example #4](https://github.com/strongbox/strongbox-authentication-example/pull/4)]_
* ~~[strongbox-web-ui](https://github.com/strongbox/strongbox-web-ui)~~ _[done in [strongbox-web-ui #2](https://github.com/strongbox/strongbox-web-ui/pull/2)]_

carlspring:
Implemented by @gokhankuyucak .

PR #515, opened 2018-01-18
#509 Updated the license information

gokhankuyucak:
Updated pom and LICENCE files


carlspring:
Thanks! It looks good!

PR #530, opened 2018-01-22
SB-1032: When a new Maven 2 repository is created, automatically create a cron task for the rebuilding and packing of indexes

fuss86:


carlspring:
Thanks! :)

PR #552, opened 2018-01-30
Fixed issues for files left behind testing

dinesh19aug:
Fixed issues for files left behind testing in MetadataHelperTest.java and MavenArtifactControllerTest.java

### DIFF
--- a/strongbox-metadata-maven-api/src/test/java/org/carlspring/strongbox/storage/metadata/MetadataHelperTest.java
+++ b/strongbox-metadata-maven-api/src/test/java/org/carlspring/strongbox/storage/metadata/MetadataHelperTest.java
@@ -10,6 +10,7 @@ import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 import java.io.*;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


strongbox-metadata-maven-api/src/test/java/org/carlspring/strongbox/storage/metadata/MetadataHelperTest.java
```diff
@@ -84,13 +85,22 @@ public static void down()
 
     private static void deleteTestResources()
     {
-        Path dirPath = Paths.get(pluginXmlFilePath).getParent();
+
+        Path dirPath = Paths.get(pluginXmlFilePath).getParent().getParent().getParent();
```
carlspring: Is this really the right way to be doing this?

carlspring:
Could you please explain:
* How you're creating the test resources
* Whether, or not it's possible to generate and keep these in memory, instead of on the file system
?

I am not sure I like this approach with the `*resource* files.

steve-todorov:
@dinesh19aug ping, any update on this? :)

strongbox-metadata-maven-api/src/test/java/org/carlspring/strongbox/storage/metadata/MetadataHelperTest.java
dinesh19aug: Since I am creating the files on the fly, I have to delete the file and folders from parent.

dinesh19aug: I couldn't find another way pf doing it

dinesh19aug:
The files are getting created on the fly. These files have to temporarily saved because the zip class can only read a file written on the disk. I am first creating an xml file, then adding an empty jar file and adding the xml file inside the jar file. Once the various test cases verifies that the xml inside the jar file matches their condition, I am deleting those jar file after the tests finish.

PR #567, opened 2018-02-03
Pr 550

SoulSilence:
sort files and dirs


strongbox-web-core/src/main/java/org/carlspring/strongbox/utils/ArtifactControllerHelper.java
```diff
@@ -225,4 +225,18 @@ else if (RepositoryFiles.isMetadata(path))
         
     }
 
+    public static List<File> getDirectories(File file) {
```
carlspring: Cosmetic: Could you please change the indentation to be in accordance with the coding convention? (Here and wherever else it the PR it's not applied).

PR #582, opened 2018-02-05
SB-1067 protect absolute paths

fuss86:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPath.java
```diff
@@ -126,8 +127,10 @@ public RepositoryPath resolve(Path other)
         {
             return this;
         }
-        
+
         other = unwrap(other);
+
+        validatePathRelativized(other);
```
sbespalov: Why did you decide to go away from the proxying target path? 
isn't it was much more reliable and interesting? 
for me it was :)

fuss86: Heh :) thanks for you interest, @sbespalov 

Yeah, I have been trying to use proxied target path, everything looked quite straightforward when I thought everything is ready. Then hundreds of tests started failing. There were many issues out there and my next "fixes" started to grow up and I started feeling that everything has become overcomplicated. A turning point that led to my decision to give up was this method: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/sun/nio/fs/WindowsPath.java#WindowsPath.toWindowsPath%28java.nio.file.Path%29 and generally `WindowsPath` doesn't want to deal with any other `Path` implementation than `WindowsPath` itself. "But hey, I want to use Proxy here and there" ;-) `ProviderMismatchException` killed my creativity :-)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPath.java
```diff
@@ -270,4 +287,30 @@ public int hashCode()
...
+        if (!PathUtils.isRelativized(target, other))
+        {
+            throw new RepositoryRelativePathConstructionException();
+        }
+    }
+
+    private void validateSibling(final Path result)
+    {
+        final Path sibling = result;
+        final String repositoryRootPath = getRoot().toString(); // String, intentionally
```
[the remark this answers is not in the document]
carlspring: Should we not `null`-proof this?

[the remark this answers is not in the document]
fuss86: it's false positive

carlspring:
Thanks! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPath.java
sbespalov: yeah, I see
you would have the same for Linux, I had to foresee this :(
sadness :(

PR #583, opened 2018-02-06
Implemented browsing in BrowseController under /browse

sanket407:
Implemented browsing under /browse.
All folders are redirected to /browse again from browsing.
All files are redirected to /storages for download.
Note: File/Artifact download not supported for Nuget currently as NugetArtifactController does not support direct file download currently.
Test this with accept-text/html


strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -63,13 +71,11 @@
     public ResponseEntity<String> browse()
     {
         try
-        {
-            Map<String,SortedSet<String>> storages = new HashMap<>();
-            storages.put("storages", new TreeSet<String>(configurationManager.getConfiguration()
-                    .getStorages()
-                    .keySet()));
+        {   
+            Path dirPath = getStoragesDir();
```
sbespalov: In the original implementation storages was fetched from configuration. 
@sanket407 have you approved with @carlspring that storages will be treated as folders now?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -97,10 +103,10 @@
                                      .body("{ 'error': 'The requested storage was not found.' }");
             }
             
-            Map<String,SortedSet<String>> repos = new HashMap<>();
-            repos.put("repositories", new TreeSet<String>(storage.getRepositories().keySet()));
+            Path dirPath = Paths.get(storage.getBasedir());
```
sbespalov: same as above.
I think that we should fetch storages and repositories form configuration.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -145,26 +151,17 @@
             
             String matchedPattern = (String) request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
             String subPath = new AntPathMatcher().extractPathWithinPattern(matchedPattern, request.getRequestURI());
```
sbespalov: take a look how `path` variable implemented in `MavenArtifactController`. You can try to use `{storageId}/{repositoryId}/{path:.+}` to expose it. I think this can suit well here.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -145,26 +151,17 @@
...
-            Path baseDir = Paths.get(repository.getBasedir().toString(), subPath);
+            Path baseDir = Paths.get(repository.getBasedir().toString(), subPath.replaceAll("/", Matcher.quoteReplacement(File.separator)));
```
sbespalov: you should resolve paths with `RepositoryProvider.resolvePath` method

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -174,4 +171,277 @@
...
+           
+            for (FileContent fileContent : content.getDirectories())
+            {
+                appendFile(sb, fileContent, requestUri, true, url);
+            }
+                        
+            if(url == null && !content.getFiles().isEmpty())
+            {
+                logger.debug("Files are present outside a repository");
+                response.setStatus(HttpStatus.INTERNAL_SERVER_ERROR.value());
```
sbespalov: can you please follow [How-To-Implement-Spring-Controllers](https://github.com/strongbox/strongbox/wiki/How-To-Implement-Spring-Controllers) guide

sbespalov:
@sanket407 can you please point me where `ArtifactResolutionService.resolveArtifactResource` method used?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
sanket407: @sbespalov I had implemented them similarly in BrowseStoragesController. There it was approved. @carlspring What do you suggest ?

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
```diff
@@ -0,0 +1,89 @@
...
+        this.directories = new ArrayList<FileContent>();
+        this.files = new ArrayList<FileContent>();
+               
+        setDirectories(dirPath);
+        setFiles(dirPath);
+    }
+       
+    public void setDirectories(Path dirPath) throws IOException
+    {   
+        Files.list(dirPath)
```
sbespalov: I think this logic should be implemented in some service instead of directly here.

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/FileContent.java
```diff
@@ -0,0 +1,61 @@
...
+{
+
+    private String name;
+
+    private String size;
+
+    private String lastModified;
+    
+    @JsonCreator
+    public FileContent( @JsonProperty("name") String name,
```
sbespalov: why you need this `@JsonProperty` here?

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/FileContent.java
```diff
@@ -0,0 +1,61 @@
...
+                        @JsonProperty("lastModified") String lastModified)
+    {
+        this.name = name;
+        this.size = size;
+        this.lastModified = lastModified;
+    }
+    
+    public FileContent(Path p)
+    {   
+        if(p == null)
```
sbespalov: same as above in `DirectoryContent`
you should populate your POJOs in service.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -174,4 +171,277 @@
...
+        Path baseDir = Paths.get(repository.getBasedir().toString(), subPath.replaceAll("/", Matcher.quoteReplacement(File.separator)));
+                
+        logger.debug("Requested browsing for {}/{}/{} ", storageId, repositoryId, subPath);
+        
+        if (repository.allowsDirectoryBrowsing() && probeForDirectoryListing(repository, subPath))
+        {
+            try
+            {
+                DirectoryContent content = new DirectoryContent(baseDir);
+                URL url = artifactResolutionService.resolveResource(storageId, repositoryId, subPath);
```
sanket407: @sbespalov  1.  Here I retrieved the URL and used it later in appendFile !

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -174,4 +171,277 @@
...
+        sb.append("<tr>");
+
+        if(isDirectory)
+        {
+            sb.append("<td><a href='" + URLEncoder.encode(name, "UTF-8") + 
+                      "/'>" + name + "/" + "</a></td>");
+        }
+        else
+        {  
+            String storagesUrl = url.toString() + "/" + URLEncoder.encode(name, "UTF-8");
```
sanket407: @sbespalov 2. Used the URL here for getting the URL pointing to /storages

sbespalov: I think `artefactUrl` will be more meaningfully instead of `storagesUrl`.
Am I understand correctly that this is URL to download artefact?

sanket407: Yes you are right!

sbespalov: also, why you concatenate `name` to URL?
this URL assumed to be ready to use without modifications

sanket407: Because URL is URL for current folder browse request  which only points to parent folder of files. So I resolve path for parent folder and append filename.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
carlspring: I believe @sbespalov has a point here. The thing is this: sure, one would normally think that clicking ".." should take you one level up, but then again... what happens if you have some sensitive directories which you've put there, (as an admin), directly on the file-system? For example, you've disabled (commented out the configuration for this in the `strongbox.xml`) some storage for the moment, or have some back up there on the file system... Since it's disabled, people shouldn't be able to browse it.
But... then you would have to the the same for repositories as well...

So, in conclusion, I believe @sbespalov is right here and we should do it as it was (and also apply this change to repositories).

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
carlspring: +1

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
carlspring: +1

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
carlspring: +1

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
carlspring: +1

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/BrowseControllerTest.java
```diff
@@ -56,13 +76,16 @@ public void testGetStorages()
...
-        assertFalse("Returned storage size does not match", returnedMap.get("storages").isEmpty());
+        DirectoryContent returned = new ObjectMapper()
+                .readValue(storages, DirectoryContent.class);
+              
+        assertNotNull("Failed to get storage list!", returned);
+        assertNotNull("Failed to get storage list!", returned.getDirectories());
+        assertFalse("Returned storage size does not match", returned.getDirectories().isEmpty());
+        
+        for(int i=0; i<returned.getDirectories().size(); i++)
+            logger.debug(returned.getDirectories().get(i).getName());
```
carlspring: Please, always use braces in cycles and conditions.

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/FileContent.java
sanket407: https://stackoverflow.com/questions/21920367/why-when-a-constructor-is-annotated-with-jsoncreator-its-arguments-must-be-ann

P.S. : This is my first time using JsonCreator.

sbespalov: ok, why do you need `JsonCreator` here? :)

sanket407: Getting foll. error without it : 
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.carlspring.strongbox.domain.DirectoryContent` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)

I think since we do not have a kind of default constructor for our POJO, the object mapper cannot create the JSON object from the String as it does not know how to. So this constructor points  the deserializer to this JsonCreator for making JSON object

sbespalov: > P.S.: this error occurred in Junit Test where I need to deserialize the string to object.

oh, now it's clear :)

sbespalov: but if you will create noags constructor and add setters then you can remove this annotations from constructor and it will look much more clearer

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
sbespalov: the main thing of `artifactResolutionService.resolveResource` method is to avoid additional modifications for URL you get from it, so I think in case of artifact files you need to construct URL just with this method without these modifications

carlspring:
@sanket407 ,

What is the state of this task?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -63,13 +79,21 @@
...
+        {               
+            Set<Entry<String, Storage>> storages = configurationManager.getConfiguration().getStorages().entrySet();
+                                                    
+            List<Path> storagePaths = new ArrayList<Path>();
+            
+            for(Entry<String, Storage> entry : storages)
+            {
+                Storage storage = entry.getValue();
+                String storageBaseDir = storage.getBasedir();
+                storagePaths.add(Paths.get(storageBaseDir));
```
sbespalov: why you are using `Paths` here again? we agreed that storages and repositories should be fetched only from configuration.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
```diff
@@ -97,10 +121,20 @@
...
+            Set<Entry<String, Repository>> repositories = storage.getRepositories().entrySet();
+            
+            List<Path> repositoryPaths = new ArrayList<Path>();
             
-            return ResponseEntity.ok(objectMapper.writer().writeValueAsString(repos));
+            for(Entry<String, Repository> entry : repositories)
+            {
+                Repository repository = entry.getValue();
+                String repositoryBaseDir = repository.getBasedir();
+                repositoryPaths.add(Paths.get(repositoryBaseDir));
```
sbespalov: you shouldn't use `Paths`, use `RepositoryProvider` instead.

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import javax.inject.Inject;
+import org.carlspring.strongbox.services.FileContentFetcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DirectoryContent
+{       
+    @Inject
+    FileContentFetcher fileContentFetcher;
```
sbespalov: I think you shouldn't inject anything here

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
sanket407: Well I have fetched storages and repositories from configuration . Then only used Paths to get the lastmodified and size from file system for ONLY THOSE repositories and storages.

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
sanket407: So I should create NEW instance of FileContentFetcher ?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
sbespalov: as I see you fetch storage content from FS here, and this is wrong
you should list repositories which defined in configuration instead

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
sbespalov: you shouldn't have `FileContentFetcher` here at all

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
sbespalov: oh, I see now what you mean 
the thing is that `fetchDirectoryContent` method name confusing a little :)

sbespalov: also, @carlspring do we need size attribute calculated for directories? in general do we need any attributes for the directories?

sanket407: @sbespalov  @carlspring  Yes I was going to mention the  same for 'size' attribute . The way I have seen on some similar listings online, they usually have a dash ( - ) in place of size for directories.

strongbox-web-core/src/main/java/org/carlspring/strongbox/domain/DirectoryContent.java
sanket407: Got you now. I thought this was the DirectoryContentFetcher .

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BrowseController.java
carlspring: @sanket407 : Yes, this seems right. You can ahead and change it.

sanket407: @carlspring  Done !

carlspring: @sbespalov : Sorry, I've just realized that this is actually a very bad idea. Every time you browse, the size of the directories will have to be re-calculated. The more your repositories grow, the slower the browsing will be.
@sanket407 : I'm very sorry for realizing this so late, but would you mind reverting this particular change?

strongbox-web-core/src/main/java/org/carlspring/strongbox/services/FileContentFetcher.java
```diff
@@ -0,0 +1,45 @@
...
+        
+        FileContent fileContent = new FileContent();
+        
+        String name = fileName.toString();
+        String size = FileUtils.byteCountToDisplaySize(path.toFile().length());
+        String lastModified = new SimpleDateFormat("dd-MM-yyyy HH-mm-ss")
+                                .format(new Date(path.toFile().lastModified()));
+        
+        fileContent.setName(name);
+        if(Files.isDirectory(path))
```
sanket407: @carlspring Here it is !

PR #588, opened 2018-02-07
protect VersionCollector from NullPointerException

fuss86:


carlspring:
Congratulations on being number 3! 🎂

Issue #572, opened 2018-02-03
The directory browsing under /storages should not contain "..", as it points to a non-existing path


carlspring:
When browsing `http://localhost:48080/storages/`, there is a link to `..` which is incorrect, as there is nothing above the path that can, or should be browseable.

The code for this is somewhere in the controllers under the `strongbox-web-core` module.

carlspring:
@sanket407 ,

It looks like this is one for you? Would you like to investigate it?

PR #592, opened 2018-02-09
Implement directory browsing for the log files 

### LINKED ISSUE #570: Implement directory browsing for the log files
We need to implement directory browsing for the log files.

There may be more the one log file in the `${STRONGBOX_VAULT}/logs` directory (please, use `PropertyUtils.getVaultDirectory()` for this).

A new method should be added to the `LoggingManagementController` that handles requests for `/logging/logs`. This should display the contents of the `${STRONGBOX_VAULT}/logs` directory.

Accessing these resources should require the user to be logged in and have the `VIEW_LOGS` privilege (defined in the `org.carlspring.strongbox.users.domain.Privileges`).

Also, please check how the directory browsing for artifacts has already been implemented in `MavenArtifactController` and follow the same approach.

adityasrini:
Fixes #570

Enabled the ability to browse the `{$STRONGBOX_VAULT}/logs` directory and any future subdirectories. Added tests to generate dummy test files via static methods and delete them after the test. Tests ensure that the test log file names and/or any sub-directories are contained in the HTML body.


adityasrini:
Issue #570 is not referenced here for some weird reason, but I am referencing it in the comment section anyway.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -245,4 +262,115 @@ public ResponseEntity uploadLogbackConfiguration(HttpServletRequest request,
...
     }
 
+    @ApiOperation(value = "Used to get log directory.",
+                  position = 0)
+    @ApiResponses(value = { @ApiResponse(code = 200,
+                                         message = "The log directory was retrieved successfully."),
+                            @ApiResponse(code = 400,
+                                         message = "Could not download log directory.") })
+    @GetMapping(value = { "/logs",
+                          "/logs/**" },
```
carlspring: Just for my better understanding, why are there two mappings for this?

adityasrini: `/logs` is to capture any URI the user entered without any trailing forward slashes. I copied this approach from `BrowseStoragesController` that has the two mappings in line 38 and the method to perform the redirect in `BaseArtifactController` in line 176. I did both in one go. If I don't do that, navigating to `/logs` gives a 404 error code, unlike `/storages`.

carlspring: Okay, thanks for the explanation! :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -238,11 +237,111 @@ public ResponseEntity uploadLogbackConfiguration(HttpServletRequest request,
...
+                                            HttpServletResponse response)
+            throws IOException
+    {
+        String requestUriString = request.getRequestURI();
+        String uriLogDirPath = requestUriString.replace("/logging", "");
+        Path localLogDirPath = Paths.get(PropertyUtils.getVaultDirectory(), uriLogDirPath);
+        
+        if (Files.notExists(localLogDirPath))
+        {
+            response.sendError(404, "File not found. Variables: " + uriLogDirPath + " " + localLogDirPath.toString());
```
steve-todorov: You need to add `return;` in this block. If you don't - the code will send the response to the client and continue executing the rest of the code.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -238,11 +237,111 @@ public ResponseEntity uploadLogbackConfiguration(HttpServletRequest request,
...
+        if (!requestUriString.endsWith("/"))
+        {
+            try
+            {
+                response.sendRedirect(requestUriString + "/");
+            }
+            catch (IOException e)
+            {
+                logger.debug("Error redirecting to " + requestUriString + "/");
+            }
```
steve-todorov: You need to watch out for the `return` after the `try/catch` block here as well. If don't `return` - it will send a `redirect` response to the client and will continue executing the rest of the code.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
adityasrini: Oh yes good catch!

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -238,11 +237,111 @@ public ResponseEntity uploadLogbackConfiguration(HttpServletRequest request,
...
+            sb.append("<h1>Index of " + uriLogDirPath + "</h1>");
+            sb.append("<table cellspacing=\"10\">");
+            sb.append("<tr>");
+            sb.append("<th>Name</th>");
+            sb.append("<th>Last modified</th>");
+            sb.append("<th>Size</th>");
+            sb.append("<th>Description</th>");
+            sb.append("</tr>");
+            sb.append("<tr>");
+            sb.append("<td colspan=4><a href=\"..\">..</a></td>");
```
bbb1991: It might produce same bug as #572 . Maybe should add something like #603 ?

adityasrini: Good note. I'll update the commit!

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
adityasrini: I tried this out and it works only if `urlPath` is mentioned. If it isn't and the user wants to just access `/logs/` a 404 will be thrown. If I try out `urlPath:.*` as a 0 or more regex match, I get a MissingParamValue exception from Spring. This seems to only work with vanilla JAX-RS when I separately tried out a `hello/{world:.*}` pattern. This was my last resort to serve the home URL and any subseqent sub URLs. Also, Spring seems to limit regex capture to the next `/` character it encounters. So if someone enters `/logs/folder1/folder2`, `{urlPath:.+}` doesn't seem to capture anything after `folder1`

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
adityasrini: Fixed in the latest commit!

Issue #590, opened 2018-02-09
Make sure AuthenticatorsConfigController is following the How-To-Write-Controllers guide


steve-todorov:
Currently the [AuthenticatorsConfigController](https://github.com/strongbox/strongbox/blob/master/strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java) is not fully following the [How-To-Write-Controllers](https://github.com/strongbox/strongbox/wiki/How-To-Implement-Spring-Controllers) guide. This needs to be fixed.

ptirador:
I'm taking this, if there's no problem.

carlspring:
Ok, great!

steve-todorov:
Fixed in #600

PR #600, opened 2018-02-11
Make sure AuthenticatorsConfigController is following the How-To-Write-Controllers guide

ptirador:
Related issue: #590 

Methods modified in UserController to follow the [controllers guide](https://github.com/strongbox/strongbox/wiki/How-To-Implement-Spring-Controllers).

The proper tests cases have been modified.

### DIFF
--- a/strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java
+++ b/strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java
@@ -8,6 +8,7 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
@@ -45,8 +46,8 @@ public class AuthenticatorsConfigController
     @ApiOperation(value = "Enumerates ordered collection of authenticators with order number and name")
     @ApiResponses(value = { @ApiResponse(code = 200, message = "The list was returned successfully.") })
     @GetMapping(value = "/",
-            consumes = MediaType.APPLICATION_JSON_VALUE,
-            produces = MediaType.APPLICATION_JSON_VALUE)
+                consumes = MediaType.APPLICATION_JSON_VALUE,
+                produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity list()
     {
         return ResponseEntity.ok(authenticatorsRegistry);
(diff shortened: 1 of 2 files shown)


strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java
```diff
@@ -28,63 +25,72 @@
...
 
     @ApiOperation(value = "Enumerates ordered collection of authenticators with order number and name")
-    @ApiResponses(value = { @ApiResponse(code = 200,
-            message = "The list was returned successfully.") })
-    @RequestMapping(value = "/", method = RequestMethod.GET, produces = { MediaType.APPLICATION_XML_VALUE,
-                                                                          MediaType.APPLICATION_JSON_VALUE })
-    public ResponseEntity<AuthenticatorsRegistry> list()
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "The list was returned successfully.") })
+    @GetMapping(value = "/",
+            consumes = MediaType.APPLICATION_JSON_VALUE,
```
steve-todorov: Cosmetic: indention should be aligned here as well as in the other methods below :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java
```diff
@@ -28,63 +25,72 @@
...
         {
             authenticatorsRegistry.reorder(first, second);
-            return ResponseEntity.ok("Reorder succeeded");
+            return getSuccessfulResponseEntity(SUCCESSFUL_REORDER, acceptHeader);
         }
         catch (Exception e)
         {
-            logger.error("Error during reorder processing.", e);
-            return toError("Error during reorder processing: " + e.getLocalizedMessage());
+            return getBadRequestResponseEntity(FAILED_REORDER, acceptHeader);
```
steve-todorov: Using `getBadRequestResponseEntity` in this particular case is maybe not a good idea because the `Exception` does not get logged in the console/logs. Could you please replace it with `getExceptionResponseEntity(HttpStatus.BAD_REQUEST, FAILED_REORDER, e, acceptHeader)`?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/AuthenticatorsConfigController.java
```diff
@@ -28,63 +25,72 @@
...
             authenticatorsScanner.scanAndReloadRegistry();
-
-            return ResponseEntity.ok("Reload succeeded");
+            return getSuccessfulResponseEntity(SUCCESSFUL_RELOAD, acceptHeader);
         }
         catch (Exception e)
         {
-            logger.error("Error during reload processing.", e);
-            return toError("Error during reload processing: " + e.getLocalizedMessage());
+            return getBadRequestResponseEntity(FAILED_RELOAD, acceptHeader);
```
steve-todorov: Same as the above comment regarding `getExceptionResponseEntity`. In addition, I believe this is one of those exceptional cases where you should actually return `Internal Server Error`, because the client won't actually be sending any data to the server and the only possible error that can occur is related to something internally handled (i.e. xml file containing the authenticators is broken or something else which is beyond the client's reach).

steve-todorov:
Everything works fine now. Many thanks for fixing this!

PR #601, opened 2018-02-11
SB-1077 HttpGetRemoteRepositoryCheckStrategy should use connection pool

fuss86:


strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.xml
```diff
@@ -49,8 +49,7 @@
                     <remote-repository url="https://repo.maven.apache.org/maven2/"
                                        download-remote-indexes="true"
                                        auto-blocking="true"
-                                       checksum-validation="true"
-                                       check-interval-seconds="20" />
```
fuss86: @carlspring I think that we should use default 60 seconds interval here too. It was added for test purposes.

strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.xml
```diff
@@ -96,8 +95,7 @@
                                        download-remote-indexes="true"
                                        auto-blocking="true"
                                        checksum-validation="true"
-                                       allows-directory-browsing="false"
-                                       check-interval-seconds="10" />
```
fuss86: as above

strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.xml
carlspring: Okay, I agree.

strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.xml
carlspring: +1.

carlspring:
Thanks! :)

PR #605, opened 2018-02-13
SB-1080 Make configuration thread-safe and make the strongbox.xml file synchronized 

fuss86:
…le on modification


strongbox-data-service/src/main/java/org/carlspring/strongbox/data/domain/GenericEntity.java
```diff
@@ -101,5 +99,9 @@ public int hashCode()
         }
         return objectId.hashCode();
     }
-    
+
+    public void copyVersionTo(final GenericEntity target) {
+        target.entityVersion = this.entityVersion;
```
sbespalov: @fuss86 just interesting, why this needed? :)

fuss86: To protect `configuration` from concurrent modification. Because of SB-930 we will be storing `String` representation of  `configuration` in the database (`BinaryConfiguration`), but we will be operating in the system on the real `Configuration` instance (once fetched from the database and deserialized). I need to keep  `BinaryConfiguration` instance versioned in the database, so e.g. once configuration was saved with `@OVersion` 2, we should not allow to save it obsolete version hold somewhere concurrently in the memory with `@OVersion` 1.

fuss86: And now it's changed to `copyTrackingFields` (see https://github.com/strongbox/strongbox/pull/605/files#diff-629f6d989f9f5c214414afe24219d1e7R103)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/BinaryConfigurationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Optional;
+
+import org.springframework.stereotype.Service;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+@Transactional
+@Service
+public class BinaryConfigurationServiceImpl
```
carlspring: I thought the configuration was stored as objects. From I am seeing, it appears that the XML is being stored in the DB? Is this the best approach? (Just checking what you think).

fuss86: I don't see a better choice right now, until we fix SB-930.

carlspring:
@fuss86 : Can I proceed with merging this, or is there still stuff to be done?

fuss86:
You could, but right now it has merge conflicts :)

fuss86:
Conflicts resolved :) Ready to be merged :)

carlspring:
Thanks! :)

PR #626, opened 2018-03-05
SB-946: Liquibase DB migration

sbespalov:


sbespalov:
@carlspring I have also checked this with distribution, and it's work

steve-todorov:
Good job @sbespalov! Thanks for fixing this! :)

PR #628, opened 2018-03-05
SB-1080 storing proper Configuration class in the database

fuss86:


strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/CustomConfiguration.java
```diff
@@ -2,13 +2,15 @@
...
 
+import javax.persistence.Entity;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlRootElement;
 
 /**
  * @author carlspring
  */
+@Entity
```
sbespalov: @fuss86 I think in this case it's better to be `@Embeddable` as you suggested instead of `@Entity` :)

fuss86: Yeah, both `@Entity` and `@Embeddable` could be then picked up by `oEntityManager.registerEntity` method. Thanks  :+1: Will change it in the next PR.

sbespalov: yeah, great :)

PR #652, opened 2018-03-19
Anonymous authorities fixed

sbespalov:


sbespalov:
@fuss86 can you please verify this PR?
The issue was related with recent configuration store changes, and it broke Anonymous authorities.
`SecurityConfig.anonymousAuthenticationFilter()` bean was created without any loaded authorities.
Maybe I was missed something and there can be other similar issues with configuration loading order?

sbespalov:
Also @fuss86 @carlspring 
I am interesting how configuration storage assumed to work now?
As I understand it drops all configuration entries from DB at application startup and then load it from XML With full rewrite of what we have in DB. Is it so?
What is the point of such approach?
Maybe it should merge XML and DB configurations somehow instead of just rewrite all in DB, or I am missing something?

fuss86:
> Maybe I was missed something and there can be other similar issues with configuration loading order?

Unfortunately I had similar `order` related issues while dealing with `configuration` handling changes. It's hard to determine this issues until they the arise (in integration tests or live). For `Anonymous authorities issue` you just hit, I think we should add one more integration test case confirming the proper bean creation order, so we will be better protected with this case for future changes.

fuss86:
> I am interesting how configuration storage assumed to work now?

The XML `configuration` file is supposed to be in-sync with database state. Every update in the database affects the XML as well. @carlspring required having up-to-date XML file so it's easy to be exported between strongbox instances.

> As I understand it drops all configuration entries from DB at application startup and then load it from XML With full rewrite of what we have in DB.

Yes, it works like this. Then strongbox instance is started based on the current XML file state. We need to ask @carlspring is it supposed to work like this, but I assumed it is.

carlspring:
>> I am interested in how configuration storage is assumed to work now?
>
> The XML `configuration` file is supposed to be in-sync with database state. Every update in the database affects the XML as well. @carlspring required having up-to-date XML file so it's easy to be exported between strongbox instances.
>> As I understand it drops all configuration entries from DB at application startup and then load it from XML With full rewrite of what we have in DB.
>
> Yes, it works like this. Then strongbox instance is started based on the current XML file state. We need to ask @carlspring is it supposed to work like this, but I assumed it is.

I was hoping we could drop the need to store the configuration in the database. @fuss86, from what I remember, the reason to keep the DB was that there could be synchronization issues...? Could we maybe use use Hazelcast to control the writes to the configuration files...?

carlspring:
@sbespalov :

I have tried deploying using `npm publish` and this branch and I now get a different exception as shown [here](https://gist.github.com/carlspring/3435a9c2a84945646eb034fc92059940).

carlspring:
Should I proceed with merging this?

sbespalov:
yes, your exception is about `npm install` now (we need to implement it to fix what you have now)

sbespalov:
> Should I proceed with merging this?

yes, please :)

sbespalov:
@fuss86 many thanks for your explanations

carlspring:
Well, then we're all good!

And no, his was during `npm publish`, not `npm install`.

PR #658, opened 2018-03-21
SB-1109: configuration resources storage fixed

sbespalov:


strongbox-data-service/src/main/java/org/carlspring/strongbox/data/domain/GenericEntity.java
```diff
@@ -59,6 +59,7 @@ public final void setObjectId(String objectId)
         this.objectId = objectId;
     }
 
+    @XmlTransient
```
sbespalov: @fuss86 should it be transient?

fuss86: Yeah, haven't noticed it before. I think XML file shouldn't have any extra information from our database so `@XmlTransient` would is expected here. Thanks.

sbespalov: ok
thanks for your response

sbespalov:
@steve-todorov now it's seems to be ready

steve-todorov:
Thanks for the fix @sbespalov! :)

PR #663, opened 2018-03-25
SB-1095 java.io.File cleanup

fuss86:


fuss86:
Branch rebased to the latest master

PR #673, opened 2018-03-30
SB-571: AQL implementation [ready]

sbespalov:


carlspring:
One thing I'd like to clarify, after having reviewed your pull request so far, is whether the keywords can be (or are) case insensitive?

Also, I think that the `LayoutProvider.getAlias()` implementation should return a single-word (case insesntive) representation so that people don't have to type in things like `Maven 2` in the exact case. Perhaps we could have a separate field for this exposed to the AQL...?

sbespalov:
> One thing I'd like to clarify, after having reviewed your pull request so far, is whether the keywords can be (or are) case insensitive?

keywords are case sensitive. we can make them case insensitive but you sure that it will be good to have `sToRaGe` valid for example?

sbespalov:
> Also, I think that the LayoutProvider.getAlias() implementation should return a single-word (case insensitive) representation so that people don't have to type in things like Maven 2 in the exact case. Perhaps we could have a separate field for this exposed to the AQL...?

ok, maybe `LayoutProvider.getAlias()` values should be renamed then if needed.

carlspring:
>> One thing I'd like to clarify, after having reviewed your pull request so far, is whether the keywords can be (or are) case insensitive?
>
>keywords are case sensitive. we can make them case insensitive but you sure that it will be good to have `sToRaGe` valid for example?

It shouldn't matter where they put upper/lowercase letters. They may have done it by accident. That shouldn't cause things to not work. Keywords in SQL are case insensitive.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/artifact/coordinates/ArtifactLayout.java
```diff
@@ -0,0 +1,16 @@
...
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target(TYPE)
+@Retention(RUNTIME)
+public @interface ArtifactLayout
```
sbespalov: @carlspring please note that I this annotation satisfies your request regarding layout keywords which can be used in AQL

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/aql/AqlController.java
```diff
@@ -0,0 +1,94 @@
...
+        Selector<ArtifactEntry> selector;
+        try
+        {
+            selector = parser.parseQuery();
+        }
+        catch (QueryParserException e)
+        {
+            // TODO: provide response error message
+
+            return ResponseEntity.badRequest().build();
```
steve-todorov: You should use the [getBadRequestResponseEntity()](https://github.com/strongbox/strongbox/blob/master/strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java#L129) to follow the convention we have introduced in the [wiki](https://github.com/strongbox/strongbox/wiki/How-To-Implement-Spring-Controllers). For the `acceptHeader` just pass `application/json`. I don't think we will be returning anything other than JSON in here for now.

sbespalov: ok. thanks @steve-todorov  :)

sbespalov: > For the acceptHeader just pass application/json. I don't think we will be returning anything other than JSON in here for now.

I assumed that it should be GET request with path variable parameter.

steve-todorov: Most of the methods in `BaseController` such as `getBadRequestResponseEntity` depend on the `acceptHeader` to be able to figure out how to return the response to the client (i.e. `text/plain` or `application/json`). However, in this case the controller will only be producing `JSON`, so you can just pass the `application/json` to `acceptHeader` :)

sbespalov:
@carlspring @steve-todorov I think this PR is mergable now.
There should be following ongoing tasks:
- case insensetive keywords;
- token values disjunction using `,`;
- dynamic grammar with layout specific keywords;

strongbox-aql/src/main/antlr4/AQL.g4
```diff
@@ -0,0 +1,241 @@
...
+or
+:
+    OR
+    | PIPE
+    | DOUBLE_PIPE
+;
+
+STORAGE
+:
+    'storage'
```
carlspring: Correct me, if I am wrong (about this keyword and the ones below), but does it mean that it doesn't support uppercase? If so, could you please add it as well (here and wherever else required)? Thanks!

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/criteria/ExpressionBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private ExpressionDialect dialect;
+
+    private Expression expression = new Expression();
+
+    public ExpressionBuilder(Class<T> targetClass,
+                             ExpressionDialect dialect)
+    {
+        super();
+        this.targetClass = targetClass;
```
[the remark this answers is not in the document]
carlspring: @sbespalov : Is this field required, or is this a false positive?

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/criteria/Predicate.java
```diff
@@ -23,6 +23,10 @@
 
     private List<Predicate> childPredicateList = new ArrayList<>();
 
+    private boolean nesteed;
```
carlspring: There's a small typo here. You mean `nested`. :)
Could you please fix the getter/setter as well? Thanks! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/artifact/coordinates/ArtifactLayout.java
carlspring: Thank you very much!

@fuss86 , @steve-todorov : You might also find this useful someday. :)

carlspring:
@sbespalov ,

Thanks for your outstanding work! This pull looks great, I've reviewed it!

I've only just seen your comment above. Yes, you're right. A really important follow-up task is to create a wiki page for this with a sufficient amount of examples! :)

I have a few really small remarks and questions, before merging it. I'm looking forward to your reply!

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/criteria/ExpressionBuilder.java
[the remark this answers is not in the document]
sbespalov: I assumed that we probably will need this field in the future.

strongbox-aql/src/main/antlr4/AQL.g4
sbespalov: I suggest to make case insensitive grammar with upcoming tasks.

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/criteria/Predicate.java
sbespalov: fixed

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/criteria/ExpressionBuilder.java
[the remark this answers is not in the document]
carlspring: Okay, just checking! :)

carlspring:
Thanks for all your hard work and patience with this! :)

steve-todorov:
Awesome!!!! Thank you for your work @sbespalov! :)

strongbox-aql/src/main/antlr4/AQL.g4
carlspring: Could you please raise a new issue for this on Github with a proper explanation?

carlspring: Could you please raise a new issue for this on Github with a proper explanation?

Issue #660, opened 2018-03-22
Refactor the UserController to use form validation and return proper responses on failed validation


steve-todorov:
The `UserController` needs to be refactored so that it uses spring form validation. Also, make sure that the controller returns the proper responses according to the [How-To-Implement-Spring-Controllers](https://github.com/strongbox/strongbox/wiki/How-To-Implement-Spring-Controllers) when there is an error in the validation.

ptirador:
I'll pick this up :)

steve-todorov:
Happy coding! :)

ptirador:
In this method:
```
public ResponseEntity updateAccessModel(@ApiParam(value = "The name of the user") @PathVariable String username,
                                        @RequestBody AccessModel accessModel,
                                        @RequestHeader(HttpHeaders.ACCEPT) String accept)
```

Do I need to validate `AccessModel` object? In that case, what should I validate? It has 3 `Maps` as attributes...

steve-todorov:
@carlspring, @fuss86 , @sbespalov could one of you pitch in and let @ptirador know what needs to be validated here. I believe the submitted `accessModel` should be validated (as well as anything that is accepted from the user and saved in the database) but I'm not sure what exactly needs to be validated.

carlspring:
@fuss86 ,

As I won't be around tonight, would you mind having a look? Many thanks! :)

fuss86:
Every map fits into the following validation rules:
* not mandatory
* if provided then
  * keys _should_ follow pattern `/storages/${storageId}/${repositoryId}`
  * values _should not_ be `empty` nor `null`

Each map is independent to the others.

ptirador:
@fuss86 Thanks for the info. What about the `User` class? It has these attributes:

```
private String username;
private String password;
private boolean enabled;
private Set<String> roles;
private String securityTokenKey;
private AccessModel accessModel;
```

steve-todorov:
`username` - required, must not be blank
`password` - required, must not be blank and must be more than 6 symbols
`enabled` - can be either true or false.
`roles` - should be an array of valid roles.
`securityToken` - this was used for `nuget` and I'm not sure, @sbespalov could you give your thoughts?
`accessModel` - I believe @fuss86 already answered this?

ptirador:
Thanks @steve-todorov!

carlspring:
@ptirador ,

Yes, you are right. There are the also these follow-up issues:
[...]

PR #677, opened 2018-04-02
Issue 660: Refactor the UserController to use form validation

ptirador:
Related issue: #660

Methods modified in `UserController` to follow the controllers guide and to use form validation.

The proper tests cases have been modified and new cases have been added.


steve-todorov:
@ptirador this looks ready for a review. Is there anything more you need to do or should I test it out? :)

ptirador:
It's allí yours!

strongbox-web-core/src/main/java/org/carlspring/strongbox/forms/users/UserForm.java
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return accessModel;
+    }
+
+    public void setAccessModel(AccessModelForm accessModel)
+    {
+        this.accessModel = accessModel;
+    }
+
+    public interface NewUser
```
carlspring: What is the the purpose of these two interfaces?

carlspring: Thanks for the explanation!

carlspring:
Thanks for these changes!

Issue #641, opened 2018-03-13
Make all invocations of the ConfigurationResourceResolver.getConfigurationResource method use property keys prefixed with "strongbox"


carlspring:
All invocations of the `ConfigurationResourceResolver.getConfigurationResource` method shouold use property keys that are prefixed with `strongbox`. For example:

```
public class AuthorizationConfigFileManager
{

    private final GenericParser<AuthorizationConfig> parser = new GenericParser<>(AuthorizationConfig.class);

    private Resource getConfigurationResource()
            throws IOException
    {
        return ConfigurationResourceResolver.getConfigurationResource("authorization.config.xml",
                                                                      "etc/conf/strongbox-authorization.xml");
    }
...
}
```

In this case:
```
return ConfigurationResourceResolver.getConfigurationResource("authorization.config.xml",
                                                              "etc/conf/strongbox-authorization.xml");
```
Should be:
```
return ConfigurationResourceResolver.getConfigurationResource("strongbox.authorization.config.xml",
                                                              "etc/conf/strongbox-authorization.xml");
```

pranavi:
I would like to work on this

carlspring:
@pranavi : Please, feel free to proceed! :-)

ptirador:
@pranavi Are you still working on this? I'm picking it up if there's no problem about it.

carlspring:
@ptirador: Please, feel free to pick this up!

PR #678, opened 2018-04-03
Issue 641: Add prefix to keys from ConfigurationResourceResolver.getConfigurationResource method.

ptirador:
Related issue: #641

All property keys from `ConfigurationResourceResolver.getConfigurationResource` have "strongbox." prefix now.


SevastyanDark:
I think it will be better that use `strongbox-` instead of `strongbox.`

carlspring:
Hi Sevastyan,

Thanks for your proposal, but -- no. It's already like this across the rest of the codebase and this is also following the typical property naming convention in Java.

ptirador:
I agree with @carlspring. In this way it's being followed the property name convention.

carlspring:
Thanks!

PR #680, opened 2018-04-05
SB-1112: problem liquibase changeset fixed

sbespalov:
@steve-todorov you probably will need to drop our production DB again because I have fixed changeset which is already applied. This changes needed as an example to create valid changesets in the future.


strongbox-data-service/src/main/resources/db/changelog/v1.0.0/v1.0.0.8__AuthorizationConfig.xml
```diff
@@ -8,6 +8,8 @@
 
         <o:dropProperty className="AuthorizationConfig" propertyName="roles" />
 
+        <o:delete tableName="Roles"/>
```
sbespalov: @fuss86 you probably need to add same statement after dropping class properties. Seems that OrientDB is not assumed to drop data within dropping a property.

steve-todorov:
Thanks for letting me know @sbespalov. Ping me again when this is merged so I can fix the instance :)

sbespalov:
@carlspring can you please review?

strongbox-data-service/src/main/resources/db/changelog/v1.0.0/v1.0.0.8__AuthorizationConfig.xml
fuss86: @sbespalov thanks for this tip. Why have you used _o_ namespace prefix here ? _delete_ element comes from liquibase core as far as I see. Will it work prefixed with _o_ namespace ?

fuss86: And how can I use it to remove all values from the property it they were not wrapped by additional element (like `Repository#groupRepositories`) ?

fuss86: Maybe we should try to use standard liquibase _update_ data change element

Issue #618, opened 2018-02-26
Upgrade the Jackson dependencies to >= 2.9.4


carlspring:
Due to vulnerabilities in our current version of Jackson, we need to upgrade to 2.9.4.

Unfortunately, just bumping the version in the `strongbox-parent/pom.xml` doesn't do the trick and it caused the following test failures:

```
[ERROR] Tests run: 26, Failures: 3, Errors: 1, Skipped: 0, Time elapsed: 0.212 s <<< FAILURE! - in org.carlspring.strongbox.controllers.users.UserControllerTest
[ERROR] testRetrieveAllUsers(org.carlspring.strongbox.controllers.users.UserControllerTest)  Time elapsed: 0.003 s  <<< ERROR!
java.lang.IllegalArgumentException: The JSON input text should neither be null nor empty.
	at org.carlspring.strongbox.controllers.users.UserControllerTest.testRetrieveAllUsers(UserControllerTest.java:209)

[ERROR] userNotExistingShouldNotUpdateAccessModel(org.carlspring.strongbox.controllers.users.UserControllerTest)  Time elapsed: 0 s  <<< FAILURE!
java.lang.AssertionError: 
1 expectation failed.
Expected status code <200> but was <500>.

	at org.carlspring.strongbox.controllers.users.UserControllerTest.getUser(UserControllerTest.java:741)
	at org.carlspring.strongbox.controllers.users.UserControllerTest.userNotExistingShouldNotUpdateAccessModel(UserControllerTest.java:706)

[ERROR] testUpdateAccessModel(org.carlspring.strongbox.controllers.users.UserControllerTest)  Time elapsed: 0 s  <<< FAILURE!
java.lang.AssertionError: 
1 expectation failed.
Expected status code <200> but was <500>.

	at org.carlspring.strongbox.controllers.users.UserControllerTest.getUser(UserControllerTest.java:741)
	at org.carlspring.strongbox.controllers.users.UserControllerTest.testUpdateAccessModel(UserControllerTest.java:664)

[ERROR] userWithoutViewUserRoleShouldBeAbleToViewHisAccountData(org.carlspring.strongbox.controllers.users.UserControllerTest)  Time elapsed: 0 s  <<< FAILURE!
java.lang.AssertionError: 
1 expectation failed.
Expected status code <200> but was <500>.

[...]

PR #690, opened 2018-04-20
Issue 618: Upgrade the Jackson dependencies to 2.9.5

ptirador:
Related issue: #618 

The `pom.xml` from parent has been modified to set the specified version.
A fix has been set for `Map` (that will be included in 2.9.6) as stated in this issue:
[Failed to specialize `Map` type during serialization where type key type incompatibility overidden via "raw" types](https://github.com/FasterXML/jackson-databind/issues/1964)


carlspring:
Thanks for this fix!

PR #691, opened 2018-04-23
SB-805: Migrate from EHCache to Hazelcast

fuss86:

PR #694, opened 2018-04-23
SB-1129: Replace the file locking with Hazelcast

carlspring:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -141,12 +151,16 @@ public void releaseLock(URI pathUri)
 
     public void acquireLock(URI pathUri)
     {
-        Lock lock = Optional.ofNullable(pathUriMap.putIfAbsent(pathUri, lock = new ReentrantLock())).orElse(lock);
+        Lock lock = Optional.ofNullable(pathUriMap.putIfAbsent(pathUri,
+                                                               lock = hazelcastInstance.getLock(pathUri.toString())))
```
carlspring: @sbespalov : Do we still need to be keeping the URI-s in the map, or should we let Hazelcast handle this?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -141,12 +151,16 @@ public void releaseLock(URI pathUri)
 
     public void acquireLock(URI pathUri)
```
carlspring: @sbespalov , @fuss86 :

This method is the only one of importance. The rest is noise, due to the fact that I based this pull request on Sergey's last pull on my branch. I will rebase once I've merged #691.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -141,12 +151,16 @@ public void releaseLock(URI pathUri)
...
     {
-        Lock lock = Optional.ofNullable(pathUriMap.putIfAbsent(pathUri, lock = new ReentrantLock())).orElse(lock);
+        Lock lock = Optional.ofNullable(pathUriMap.putIfAbsent(pathUri,
+                                                               lock = hazelcastInstance.getLock(pathUri.toString())))
+                            .orElse(lock);
         lock.lock();
 
         while (!pathUriMap.containsKey(pathUri))
```
carlspring: @sbespalov : I know I raised this in the chat, but I think it would be great, if you could explain the purpose of this `while` loop here, as I'm not sure I'm getting it.

sbespalov: this loop used to here to avoid "race donditions". For example two threads A and B can get same lock object, and then A accure lock first and remove this lock object from Map with `ArtifactManagementService.releaseLock`, so thread B will lock it after but other threads will not see this because there will be no this lock object in Map any more. So with this loop we guarantie that if thread accured lock then this lock object contains in Map and other threads can see this.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
sbespalov: yeah, we don't need this Map with Hazelcast because Hazelcats will be this Map :)
but we should think how to invalidate `ILock` objects like this was done in `ArtifactManagementService.releaseLock` to avoid memory leaks. I think we should use `ILock.destroy()` somehow, maybe there is some examples of how to do this correctly in concurrent environment.

fuss86:
@sbespalov, could you review these changes, please ?

sbespalov:
@carlspring @fuss86 maybe we should add some UnitTests to check concurrent behaviour? For example such test can deploy same artifact in loop within 2 threads in parallel.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPathLock.java
```diff
@@ -0,0 +1,41 @@
...
+    @Inject
+    private HazelcastInstance hazelcastInstance;
+
+    /**
+     * @see <a href="http://docs.hazelcast.org/docs/latest-development/manual/html/Distributed_Data_Structures/Lock.html#page_ILock+vs.+IMap.lock">ILock vs. IMap.lock</a>
+     */
+    public void lock(RepositoryPath repositoryPath)
+    {
+        final IMap mapLocks = hazelcastInstance.getMap(REPOSITORY_PATH_MAP_LOCKS);
+        final URI lock = repositoryPath.toUri();
```
sbespalov: One thing I suggest to add here. We need to assert `URI.isAbsolute()`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPathLock.java
```diff
@@ -0,0 +1,41 @@
...
+        // DEV NOTE: map may be empty for locking
+        // we don't have to put values to the map
+        mapLocks.lock(lock);
+    }
+
+    public void unlock(RepositoryPath repositoryPath)
+    {
+        final IMap mapLocks = hazelcastInstance.getMap(REPOSITORY_PATH_MAP_LOCKS);
+        final URI lock = repositoryPath.toUri();
```
sbespalov: same as above

fuss86:
> @carlspring @fuss86 maybe we should add some UnitTests to check concurrent behaviour? 

Test added, @sbespalov , @carlspring

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPathLock.java
fuss86: done

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPathLock.java
fuss86: done

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/resources/artifacts/lorem-ipsum-1.txt
```diff
@@ -0,0 +1,39 @@
+Eĥo pleje liternomo hu un', si danke kilometro mallongigoj sia. Ehe it identiga parentezo. Ko pra rolfinaĵo malprofitanto, kaŭzo prepoziciaĵo veo u, jo orda neigi estiel sub. Eko gardi kunmetita da, kio jh kiel aligi nuancilo. Sis ig tempa hebrea. Us unu afrikato anstataŭi, danki aperi aspektismo et nun, it tuta tiom traigi cit. Hekto multe ci ism, tri nj mekao esperanteca.
```
carlspring: I believe it would be better to just use `new RandomInputStream(length)` and read from it, instead of having these files.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
```diff
@@ -482,4 +509,62 @@ public void testRemoveTimestampedSnapshots()
...
+        {
+            this.repositoryPath = repositoryPath;
+            this.is = is;
+        }
+
+        @Override
+        public Void call()
+                throws Exception
+        {
+            mavenArtifactManagementService.store(repositoryPath, is);
```
sbespalov: @fuss86 we probably need to check that there were no exceptions here.

carlspring:
Thanks! :)

PR #699, opened 2018-05-03
SB-1131 Upgrade orientDb to 3.X.X [in progress]

fuss86:


fuss86:
@sbespalov could you please review this PR ?

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
```diff
@@ -68,14 +94,14 @@ private EntityManager createEntityManager(final OJPAProperties properties)
     @Override
     public void close()
     {
-        databasePool.close();
+        dataSource.close();
```
sbespalov: @fuss86 are you sure that we heed to close all `dataSoruce` here? I mean that it will close underling `OrientDB` judging by the source code.

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
```diff
@@ -1,31 +1,45 @@
...
 
-    public OJPAPartitionedEntityManagerPool(final OJPAProperties properties)
+
+    public OJPAObjectDatabaseTxEntityManagerFactory(final OJPAProperties properties,
+                                                    final OrientDataSource dataSource)
     {
         this.properties = properties;
-        this.databasePool = new OPartitionedDatabasePool(properties.getURL(), properties.getUser(),
-                properties.getPassword(), 100, 100);
+        this.dataSource = dataSource;
```
sbespalov: @fuss86 what is the difference between `OPartitionedDatabasePool` and `OrientDataSource`? I mean that this guys  seems created too many different `Pool` implementations and I am confused :)

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
```diff
@@ -107,15 +78,16 @@ public EntityManagerFactory entityManagerFactory()
 
         LocalContainerEntityManagerFactoryBean result = new LocalContainerEntityManagerFactoryBean();
         result.setJpaPropertyMap(jpaProperties);
-        result.afterPropertiesSet();
+        result.setDataSource(dataSource);
+        result.setPersistenceUnitName("strongbox-PU");
```
sbespalov: shouldn't this and below be defined in `persistence.xml`?

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
```diff
@@ -107,15 +78,16 @@ public EntityManagerFactory entityManagerFactory()
...
+        result.setPackagesToScan("org.carlspring.strongbox");
+        result.setPersistenceProvider(new OJPAObjectDatabaseTxPersistenceProvider());
 
-        return result.getObject();
-    }
+        // @sbespalov, needed ?
+        // ServiceLoader.load(OSQLFunctionFactory.class);
```
sbespalov: if it works without it then I think it don't needed any more :)

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
```diff
@@ -57,48 +47,29 @@
...
-
-
+        return liquibase;
     }
 
     @Bean
-    public PlatformTransactionManager transactionManager()
+    public PlatformTransactionManager transactionManager(EntityManagerFactory emf)
     {
-        return new JpaTransactionManager(entityManagerFactory());
```
sbespalov: @fuss86 can't we leave this as it was without `@Bean` method arguments (everywhere in this configuration)? Once I found that Spring better resolves circular/transitive bean dependencies with such approach.

sbespalov:
@fuss86 this PR looks very good to me, thanks. I just have some questions/suggestions. Also I was made custom JPA implementation some time ago because as I understand that time they have no connection pool with original JPA implementation. So is it still the same situation? If so maybe we really should give it back to them?

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
fuss86: Yes, I believe so it's absolutely OK here. 

https://docs.oracle.com/html/E13946_03/ejb3_overview_emfactory_close.html

`EntityManagerFactory.close` will be invoked when the application shutdown has been invoked.

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
fuss86: >  I mean that this guys seems created too many different Pool implementations and I am confused :)

I'm really confused too :( 

The bad thing is in v. 3.0.0 [OPartitionedDatabasePool#acquire](https://github.com/orientechnologies/orientdb/blob/3.0.0/core/src/main/java/com/orientechnologies/orient/core/db/OPartitionedDatabasePool.java#L176) returns [deprecated ODatabaseDocumentTx](https://github.com/orientechnologies/orientdb/blob/3.0.0/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java) instance with comment 

> @Deprecated use {@link OrientDB} instead.

So I started using [OrientDB](https://github.com/orientechnologies/orientdb/blob/3.0.0/core/src/main/java/com/orientechnologies/orient/core/db/OrientDB.java) wherever needed (at the end it showed up that I had to follow this implementation over all the code, right next to [ODatabasePool](https://github.com/orientechnologies/orientdb/blob/3.0.0/core/src/main/java/com/orientechnologies/orient/core/db/ODatabasePool.java) and [OrientDataSource](https://github.com/orientechnologies/orientdb/blob/3.0.0/jdbc/src/main/java/com/orientechnologies/orient/jdbc/OrientDataSource.java)).

OrientDb docs for v.3.0.x [suggests to use ODatabasePool](https://orientdb.com/docs/3.0.x/java/Document-API-Database.html#using-database-pools). I have been digging in their code a while and it looks like they are changing their API a bit but old implementation is still around.

fuss86: > I mean than we need to be sure that it will always provide only one connection within current transaction.

We're providing [only one connection](https://github.com/fuss86/strongbox/blob/e79a791497e5911c67ae450b80afed56d5163db4/strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java#L78) on entity manager creation, why do you suspect more than one connection ?

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
fuss86: We can drop the xml (persistence.xml precisely) and configure persistence unit programmatically this way by using Java config and Spring features. I prefer manipulating java code than xml.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
fuss86: Let's check ;)

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
fuss86: If you deliver `@Bean` method argument like in this example: `PlatformTransactionManager transactionManager(EntityManagerFactory emf)` you're expecting `EntityManagerFactory` initialized instance to actually initialize `PlatformTransactionManager` so it's the analogous way and clearer a bit in my opinion.

fuss86:
> So is it still the same situation? If so maybe we really should give it back to them?

I believe so. AFAIU they have changed the core API to manipulate databases and pools.

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
sbespalov: oh, seems that I mixed it up with the `EntityManager`
you are certainly right of course, thanks :)

strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManagerFactory.java
sbespalov: yes, I mean that we need to be sure that we have only one connection per transaction.

thanks for explanations :)

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
sbespalov: yeah, seems that you reworked the configuration structure, so maybe you are right. I mean that we should resolve bean dependencies with expectedly configuration method calls if possible rather then method arguments injection.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/DataServiceConfig.java
carlspring: +1: I agree. Wherever possible, we should do the configuration programatically.

fuss86: Yes, it is removed in this PR :)

strongbox-parent/pom.xml
```diff
@@ -92,7 +92,7 @@
         <version.hibernate.validator>6.0.7.Final</version.hibernate.validator>
         <version.glassfish.javax.el>3.0.1-b08</version.glassfish.javax.el>
         <version.antlr>4.7.1</version.antlr>
-        <version.liquibase>3.5.5</version.liquibase>
+        <version.liquibase>3.5.3</version.liquibase>
```
carlspring: @fuss86 : Just for our better understanding (and for future reference) -- what was the reason this had to be rolled back to an older version?

Issue #657, opened 2018-03-20
Create and store the history of executed scheduled jobs


fuss86:
Please create and store the history of executed scheduled jobs. 

This will require creation of new `@Entity`/`GenericEntity` classes (see `org.carlspring.strongbox.domain.ArtifactEntry` as an example) and new liquibase changeset (see `v1.0.0.2__ArtefactEntry.xml`) as an example.

The history should contain a list of entries with:
* task name
* thread name
* start datetime
* end datetime (either fail or success)
* status (started, fail, success)
* exception message

adityasrini:
Sign me up!

steve-todorov:
@adityasrini ping, have you had a chance to work on this yet? :)

adityasrini:
Not yet since I was involved with some csv to database sanitization stuff! I'll pick this up now and ping you on gitter.

divyad1508:
Hi @fuss86 / @carlspring .

I am thinking of picking up this issue. Please can you help me understanding the issue better.

Thanks!

evsikov:
@carlspring, can I work on this one?

fuss86:
I'm sure @carlspring would say `yes` :) Pick it up :)

carlspring:
Of course! Please, go ahead!

evsikov:
On it.

carlspring:
You might want to look at #1373 as well. To be honest, I'm not sure, if that's not a duplicate (@fuss86 , @sbespalov : could you please confirm, thanks!).

sbespalov:
#1373 is not dup

carlspring:
Okay, thanks!

PR #701, opened 2018-05-07
Create and store in the database the history of executed scheduled jobs

adityasrini:
Fixes issue #657


steve-todorov:
@sbespalov could you please have a look at this and let @adityasrini know if you have any remarks?

sbespalov:
@adityasrini is this PS assumed to be complete? I am asking becasue I can't find any logic which manages  `ScheduledTaskEntry`.

strongbox-cron/strongbox-cron-api/src/main/resources/db.changelog/v1.0.0/v1.0.0.11__ScheduledTaskEntry.xml
```diff
@@ -0,0 +1,23 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
```
sbespalov: changelog files should be placed in `data-service-config` module

carlspring:
@adityasrini .

What is the status of this PR?

adityasrini:
I still have more to go. I haven't added logic to handle `ScheduledTaskEntry` yet and since I need some guidance on that so I PM'ed sbespalov.

adityasrini:
@sbespalov, I have a quick question. Is this how I should design and implement the ScheduledTaskEntryService interface?
https://gist.github.com/adityasrini/e7bb95fb4fa805c7964f118bb189f04b

adityasrini:
Cross posting a comment I left on gist since gist doesn't send any notifications for activity.

 > Yeah I don’t think I’d need all. I was just going by how ArtifactEntry services were implemented which includes a bunch of methods that make sense for retrieving artifacts and not necessarily scheduled tasks. Just wanted to clarify this for that reason!

adityasrini:
@sbespalov is it safe to assume that the task names have a one to many relationship with thread names? Like one task can run on many threads.

adityasrini:
Because if there is a one to many relationship anywhere, I can implement the necessary methods to handle that and use a `Set` instead of a `String` for thread name.

sbespalov:
@adityasrini I think you don't need any Entity relations here, `ScheduledTaskEntry` probably should be just plane entry with only thread name.

sorry for late response :)

andreybleme:
Hi, @adityasrini and @sbespalov what is avoiding this code to be merged?

andreybleme:
@carlspring I'll wait for @adityasrini 's response :)

adityasrini:
@andreybleme have at it :)

adityasrini:
@carlspring shall I close this PR since @andreybleme will be picking this up?

carlspring:
Hi @adityasrini ,

Thanks for replying! Yeah, I think it would be best to close and restart it from scratch. Thanks for effort!

PR #706, opened 2018-05-18
SB-1137: `ArtifactCoordinates` fixed for `RemoteArtifactEntiry`

sbespalov:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/AbstractRepositoryProvider.java
```diff
@@ -140,14 +149,15 @@ protected RepositoryInputStream decorate(String storageId,
...
     {
         Assert.isInstanceOf(RepositoryPath.class, path);
-        return getOutputStream((RepositoryPath) path);
+        OutputStream os = getOutputStreamInternal((RepositoryPath) path);
+        
+        return decorate((RepositoryPath) path, os);
     }
     
-    protected abstract RepositoryOutputStream getOutputStream(RepositoryPath repositoryPath)
+    protected abstract OutputStream getOutputStreamInternal(RepositoryPath repositoryPath)
```
carlspring: Why is it necessary to rename these methods?

sbespalov: because there is two `getOutputStream` methods and it causes a confusions and ambiguity, some tests was failed because wrong `getOutputStream` method was invoked within test artifacts generation.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/AbstractRepositoryProvider.java
```diff
@@ -158,25 +162,31 @@ protected final RepositoryOutputStream decorate(RepositoryPath repositoryPath,
...
-    public void onBeforeWrite(RepositoryStreamContext ctx)
+    public void onBeforeWrite(RepositoryStreamContext ctx) throws IOException
     {
-        logger.debug(String.format("Writing [%s]", ctx.getPath()));
+        RepositoryPath repositoryPath = (RepositoryPath) ctx.getPath();
+        String path = RepositoryFiles.stringValue(repositoryPath);
+        
+        logger.debug(String.format("Writing [%s]", path));
+        
+        if (!RepositoryFiles.isArtifact(repositoryPath)) {
```
sbespalov: @carlspring @fuss86 you should note that only Artifacts stored in DB now, as discussed within `SB-984`. But also we need make `pom.xml` to be artifacts and have `ArtifactCoordinates `within upcoming tasks.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/AbstractRepositoryProvider.java
carlspring: I see. Thanks for explaining! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/AbstractRepositoryProvider.java
carlspring: +1: Could you please raise a follow-up task for this?

carlspring:
Thanks!

PR #707, opened 2018-05-20
Issue 408: Upgrade org.apache.httpcomponents:* to >= 4.5.5

ptirador:
Related issue: #408 

Updated `pom.xml` from `strongbox-parent`.

### DIFF
--- a/strongbox-parent/pom.xml
+++ b/strongbox-parent/pom.xml
@@ -65,7 +65,7 @@
         <version.guava>24.0-jre</version.guava>
         <version.hamcrest>1.3</version.hamcrest>
         <version.hazelcast>3.9.3</version.hazelcast>
-        <version.httpclient>4.5</version.httpclient>
+        <version.httpclient>4.5.5</version.httpclient>
         <version.jackson>2.9.5</version.jackson>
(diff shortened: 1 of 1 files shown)


carlspring:
@ptirador ,

Have you checked whether the issue described in #451 is still happening? BTW, please note, that [Jersey](https://jersey.github.io/) is now [becoming part of the Eclipse Foundation](http://blog.supol.info/?p=13). If https://github.com/jersey/jersey/issues/3629 has been fixed (and it's not clear, whether or not it was included as a fix in [2.27](http://blog.supol.info/?p=28) which has been released), then we should be okay. Could you please check? Does this impact the build time, etc?

carlspring:
@ptirador ,

I've tried your PR and it looks good to me.

@fuss86 , @sbespalov :

If you guys experience any odd behaviour related to the HTTP client/pool, this might be the cause.

carlspring:
@ptirador ,

I've reverted this in 0a234a936132706907237ed0ad7d868d2f77a983, as it seems to be breaking the `strongbox-web-integration-tests` for the `maven` sub-module. If you have some time, could you try investigating this further? Thanks! :)

PR #708, opened 2018-05-21
SB-1119 performance improvements [in progress]

fuss86:


strongbox-data-service/src/main/java/com/orientechnologies/orient/object/jpa/OJPAObjectDatabaseTxEntityManager.java
```diff
@@ -414,7 +414,7 @@ public EntityTransaction getTransaction()
     @Override
     public void close()
     {
-        database.close();
+        //database.close();
```
fuss86: `ODatabaseException: The database instance is not set in the current thread.`

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/service/CommonCrudService.java
```diff
@@ -49,7 +59,7 @@ public void postConstruct()
     {
         identifyEntity(entity);
         
-        ReflectionUtils.doWithFields(entity.getClass(), (field) -> {
+        ReflectionUtils.doWithFields(this.getEntityClass(), (field) -> {
```
sbespalov: I think we can try to drop this `getEntityClass()` method and use some utility which will expose target class from underlying `InvocationHandler`.

carlspring:
@fuss86,

Is this now obsolete?

fuss86:
> Is this now obsolete?

Please leave it as it is for the time being :)

fuss86:
Fixed in other PRs.

PR #709, opened 2018-05-22
SB-1138: Implement tests for raw proxy/group repositories

carlspring:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/AbstractRepositoryProvider.java
```diff
@@ -246,10 +239,18 @@ public void onBeforeRead(RepositoryStreamContext ctx) throws IOException
 
         ArtifactEntry artifactEntry = provideArtirfactEntry(storageId, repositoryId, path);
 
+        /*
+        // TODO: @sbespalov:
+        // TODO: This is breaking the raw layout provider.
+        // TODO: Any advice on how to fix it?
+        // TODO: Can we remove this check permanently?
+
```
carlspring: @sbespalov , @fuss86 ,

Any thoughts, or advice here?

fuss86: I assume you're asking about 

> if (!RepositoryFiles.isArtifact(repositoryPath))

If so, without it, our database will contain a metadata/ArtifactEntry about every possible read/write file from/to our system, as far as I understand

carlspring:
Thanks for the advice and help, guys! :)

PR #713, opened 2018-05-25
SB-1087: Embedded OrientDB Studio

sbespalov:


sbespalov:
@carlspring I will resolve Sonar issue a bit later today.

carlspring:
This is great! Thank you!

* We need to limit the access to this. Some options:
  * Would it be possible to secure this with Spring Security?
  * If not, maybe we could have a property like `strongbox.orientdb.studio.enabled`?
* Could please double-check under what license the OrientDB Studio is?

Thanks!

sbespalov:
> Could please double-check under what license the OrientDB Studio is?
[it's Apache 2.0](https://github.com/orientechnologies/orientdb-studio/blob/master/license.txt)

sbespalov:
> Would it be possible to secure this with Spring Security?

No, it using embedded server and different port.

sbespalov:
> If not, maybe we could have a property like strongbox.orientdb.studio.enabled?

ok, will make such property

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/ConnectionConfigOrientDB.java
```diff
@@ -15,7 +15,8 @@
     private static final Logger logger = LoggerFactory.getLogger(ConnectionConfigOrientDB.class);
 
     public static final String PROPERTY_PROFILE = "strongbox.orientdb.profile";
-
+    public static final String PROPERTY_STUDIO_ENABLED = "strongbox.orientdb.studio.enabled";
```
sbespalov: @carlspring you can disable Studio with `-Dstrongbox.orientdb.studio.enabled=false`. By default it's enabled.

carlspring: Actually, I think it would probably be better to have this disabled by default. The reason is this -- this is not an application that we, ourselves, are developing. If it has vulnerabilities, we'll most-likely  find out about them a bit late. We wouldn't like to expose our users to an unnecessary threat. This functionality would be useful for the following things:
- Development
-  Debugging
- Analyzing issues

@fuss86 , @steve-todorov : What do you guys think?

We could have this enabled in our projects during development (in the `strongbox-web-core` and `strongbox-distribution`).

sbespalov: @carlspring we can have it explitely disabled in Jetty distribution, but it will be enabled by default for devalopment.

carlspring:
>> If not, maybe we could have a property like strongbox.orientdb.studio.enabled?
> 
> ok, will make such property

Could you please also update the [Application Properties](https://github.com/strongbox/strongbox/wiki/Application-Properties) page when you're done? Thanks! :)

strongbox-data-service/pom.xml
```diff
@@ -276,7 +276,12 @@
             <groupId>org.reflections</groupId>
             <artifactId>reflections</artifactId>
         </dependency>
-                
+
+        <dependency>            
+            <groupId>org.webjars</groupId>
+            <artifactId>orientdb-studio</artifactId>
```
carlspring: Could you please clarify how [^this^](https://github.com/webjars/orientdb-studio) differs from [OrientDB's project](https://github.com/orientechnologies/orientdb-studio)?

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -55,18 +73,117 @@ public void start()
         }
     }
 
+    public File getStudioClasspathLocation()
+    {
+        return new File(OServer.class.getProtectionDomain()
+                                     .getCodeSource()
+                                     .getLocation()
+                                     .getPath()).toPath().resolveSibling("orientdb-studio-2.2.0.jar").toFile();
```
carlspring: This code would have to keep being updated every time there is a newer version of the library. Would it be possible to improve this so that it could be a bit more dynamic and future-proof?

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -55,18 +73,117 @@ public void start()
...
+                JarEntry file = enumEntries.nextElement();
+                if (!file.getName().startsWith("META-INF/resources/webjars/orientdb-studio/2.2.0/"))
+                {
+                    continue;
+                }
+                
+                Path filePath = studioPath.resolve(file.getName()
+                                                       .replace("META-INF/resources/webjars/orientdb-studio/2.2.0/",
+                                                                ""));
+                if (file.isDirectory()) {
```
carlspring: Cosmetic: Curly braces on new line, please.

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -55,18 +73,117 @@ public void start()
...
+        logger.info(String.format("Initialize OrientDB Studio at [%s].", studioPath.toAbsolutePath().toString()));
+        Files.createDirectories(studioPath);
+        
+        try (JarFile jar = new JarFile(studioClasspathLocation))
+        {
+            Enumeration<JarEntry> enumEntries = jar.entries();
+            while (enumEntries.hasMoreElements())
+            {
+                JarEntry file = enumEntries.nextElement();
+                if (!file.getName().startsWith("META-INF/resources/webjars/orientdb-studio/2.2.0/"))
```
carlspring: Same as my earlier comment about trying to do it a bit more dynamically...

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -81,23 +198,26 @@ private void init()
...
         System.setProperty("ORIENTDB_ROOT_PASSWORD", connectionConfig.getUsername());
 
         // add other properties
         List<OServerEntryConfiguration> properties = new LinkedList<>();
         properties.add(buildProperty("server.database.path", getDatabasePath()));
         properties.add(buildProperty("plugin.dynamic", "false"));
-        properties.add(buildProperty("log.console.level", "info"));
-        properties.add(buildProperty("log.file.level", "fine"));
+        properties.add(buildProperty("log.console.level", "linfo"));
+        //properties.add(buildProperty("log.file.level", "fine"));
```
carlspring: If this is not required, could you please remove it?

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -81,23 +198,26 @@ private void init()
...
+
         System.setProperty("ORIENTDB_ROOT_PASSWORD", connectionConfig.getUsername());
 
         // add other properties
         List<OServerEntryConfiguration> properties = new LinkedList<>();
         properties.add(buildProperty("server.database.path", getDatabasePath()));
         properties.add(buildProperty("plugin.dynamic", "false"));
-        properties.add(buildProperty("log.console.level", "info"));
-        properties.add(buildProperty("log.file.level", "fine"));
+        properties.add(buildProperty("log.console.level", "linfo"));
```
carlspring: Is this a typo? (`linfo`)?

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
```diff
@@ -141,6 +261,12 @@ private String getDatabasePath()
         return getVaultDirectory() + "/db";
     }
 
+    private String getStudioPath()
+    {
+        return getVaultDirectory() + "/www";
```
carlspring: Could we call this something more like... `orientdb-studio`...?

carlspring:
This pull looks great, thanks!

I have some small remarks, questions and suggestions.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/ConnectionConfigOrientDB.java
steve-todorov: Yes, these are valid points. Maybe we should have this enabled only on demand whenever a specific property is passed.

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
sbespalov: fixed

strongbox-data-service/pom.xml
sbespalov: `org.webjars:orientdb-studio` is just [webjars.org](https://www.webjars.org/) packaged **OreitntDB Studio**.  Unfortunately I was not able to make it work using original studio distribution from  `'com.orientechnologies:orientdb-studio:jar:3.0.1'` so I used it from **webjars**.

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
sbespalov: The thing is that we need to know OrientDB jar file name and this was the only way I found. Also I made a constant with the version for it to be more convenience to update versions.

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
sbespalov: fixed

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
sbespalov: fixed

strongbox-data-service/src/main/java/org/carlspring/strongbox/data/server/EmbeddedOrientDbServer.java
sbespalov: It called so in OrietntDB so I think we should follow this conversion.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/ConnectionConfigOrientDB.java
sbespalov: @carlspring @steve-todorov regarding our discussion in the channel, I confirm that in the future we can make "proxy" controller which will pass Spring managed requests into Studio.

carlspring: @sbespalov ,

Okay! Would you mind raising a follow-up task for this, please?

sbespalov: yes, will do it tomorrow

carlspring:
Thanks, Sergey! :)

PR #732, opened 2018-06-22
SB-1164: storage/repository browsing improvements

sbespalov:


strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -260,7 +265,8 @@ public Object generateLogDirectoryListing(@PathVariable("urlPath") Optional<Stri
                 return getBadRequestResponseEntity("Requested path is not a directory!", acceptHeader);
             }
 
-            DirectoryListing directoryListing = DirectoryListing.fromPath(logsBaseDir, requestedLogPath);
+            //XXX: make logs folder to be a Raw Repository!!!
```
sbespalov: @carlspring I think this will be great feature!
what do you think?

strongbox-web-core/src/main/java/org/carlspring/strongbox/services/DirectoryListingServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+            fileContent.setRepositoryId(repository.getId());
+            
+            fileContent.setUrl(calculateDirectoryUrl(fileContent));
+        }
+
+        return directoryListing;
+    }
+
+    @Override
+    public DirectoryListing fromPath(RepositoryPath path)
```
steve-todorov: I believe we should also preserve the previous implementation here `public DirectoryListing fromPath(Path rootPath, Path path)`. The path you might want to generate `DirectoryListing` for might actually be some FS path on the server outside the context of `RepositoryPath`, `Storage`, and `Repository` (i.e. like the browsing the logs or something else which might need this in the future).

sbespalov: `RepositoryPath`  have this under the hood

steve-todorov:
Great work @sbespalov! Thanks for fixing this! :) 

@carlspring we can merge this now if you don't have any remarks.

carlspring:
Thank you very much, @sbespalov! :)

PR #734, opened 2018-06-26
SB-1167: `StrongboxConfigurationController` json content type fixed

sbespalov:
- StrongboxConfigurationController moved under `/api/configuration/strongbox` with multi content support


carlspring:
Thanks! :)

PR #736, opened 2018-06-26
strongbox-web-integration-tests hotfix [Nuget and Maven]

fuss86:


fuss86:
use #737 instead of this one

PR #744, opened 2018-07-06
SB-1174: Async events

sbespalov:


strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
```diff
@@ -0,0 +1,130 @@
...
+        logger.info("Jetty environment detected.");
+
+        return executor;
+    }
+
+    private Executor lookupTomcatExecutor()
+    {
+        try
+        {
+            Class.forName("org.apache.catalina.core.ApplicationContextFacade");
```
carlspring: Isn't it possible to get the `executor` from the `servletContext` in a similar way to how you're doing it in the `lookupJettyExecutor` method?

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventTypeEnum.java
```diff
@@ -33,11 +33,6 @@
      */
     EVENT_ARTIFACT_CHECKSUM_UPDATED(5),
 
-    /**
-     * Occurs when an artifact metadata file has been uploaded.
-     */
-    EVENT_ARTIFACT_METADATA_UPLOADED(6),
```
carlspring: What's the reason to remove this?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/config/NugetLayoutProviderConfig.java
```diff
@@ -22,12 +22,10 @@
 
 @Configuration
 @ComponentScan({ "org.carlspring.strongbox.configuration",
-                 "org.carlspring.strongbox.event",
```
carlspring: (Just for my better understanding) : You've removed this, because it's handled via Spring's Event API, right?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-raw-layout-provider/src/test/java/org/carlspring/strongbox/config/RawLayoutProviderTestConfig.java
```diff
@@ -8,6 +8,7 @@
  */
 @Configuration
 @Import({ RawLayoutProviderConfig.class,
+          EventsConfig.class,
```
carlspring: ... or not? :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactUpdatedEventListener.java
```diff
@@ -14,7 +15,8 @@
         extends BaseMavenArtifactEventListener
 {
 
-    @Override
+    @EventListener
```
fuss86: `@EventListener` not needed

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/storage/metadata/MavenMetadataManager.java
```diff
@@ -420,21 +418,19 @@ public void mergeAndStore(final Path metadataBasePath,
 
     }
 
-    private void doInLock(final Path metadataBasePath,
-                          final Consumer<Path> operation)
+    private void doInLock(RepositoryPath metadataBasePath,
+                          Consumer<Path> operation)
     {
-        synchronized (getMetadataSynchronizationLock(metadataBasePath))
+        repositoryPathLock.lock(metadataBasePath);
```
fuss86: :+1:

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.catalina.Service;
+import org.apache.catalina.core.ApplicationContext;
+import org.apache.catalina.core.ApplicationContextFacade;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.FactoryBean;
+import org.springframework.core.task.SyncTaskExecutor;
+import org.springframework.data.util.ReflectionUtils;
+
+public class EventExecutorFactoryBean implements FactoryBean<Executor>
```
fuss86: why can't we use default `org.springframework.core.task.SimpleAsyncTaskExecutor` or even custom `java.util.concurrent.ThreadPoolExecutor` ? I don't like it it's container specific solution

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -153,9 +155,13 @@ public long store(RepositoryPath repositoryPath,
         {
             repositoryPathLock.unlock(repositoryPath);
         }
+        
+        result.getEvents().stream().forEach(Runnable::run);
```
fuss86: ` result.getEvents().stream().forEach(Runnable::run);`

we should use ` @TransactionalEventListener` instead of this custom workarounds

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
sbespalov: seems no, at least I was not able to find alternatives

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/artifact/ArtifactEventTypeEnum.java
sbespalov: because it not used, and many other events not used, almost all of them
why do we need unused events?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/main/java/org/carlspring/strongbox/config/NugetLayoutProviderConfig.java
sbespalov: yes, you are right

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-raw-layout-provider/src/test/java/org/carlspring/strongbox/config/RawLayoutProviderTestConfig.java
sbespalov: I can't say for sure is it cleanup or not :)
but you should know that we have a real problem with Spring configuration classes and @CompontntScan. The problem is that all our modules share same package names like `org.carlspring.strongbox.services` or `org.carlspring.strongbox.config` which cause and this ambiguous behavior of @CompontntScan annotation. We can't say for sure what configuration will fetch beans from such packages and how to manage bean dependencies correctly.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/Event.java
```diff
@@ -1,22 +1,22 @@
 package org.carlspring.strongbox.event;
 
+import org.springframework.context.ApplicationEvent;
+
 /**
  * @author carlspring
  */
-public class Event
+public class Event extends ApplicationEvent
```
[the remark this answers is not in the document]
sbespalov: yeah, I have tried without `ApplicationEvent` but it was not worked. I will be pleasure if you can advice what was the problem.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/IndexedMavenArtifactStoredEventListener.java
```diff
@@ -24,7 +26,8 @@
     @Inject
     private MavenIndexGroupRepositoryComponent mavenIndexGroupRepositoryComponent;
 
-    @Override
+    @EventListener
```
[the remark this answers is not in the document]
sbespalov: removed, thanks

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactUpdatedEventListener.java
sbespalov: removed, thanks

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
sbespalov: the thing is that if we produce Async events within request thread and then execute them using another executor this can cause Event queue overflowing. 

I have tried it with limit of 100 Events in queue and this was not enought even for our strongbox-against-strongbox. In real production life there will be much more events. 


After understanding of this problem I came with this solution to share same executor for HTTP Requests and Async Events, which will bring correlation for events and requests and we will not be able to produce more events then we can execute.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
sbespalov: good suggestion, I like it
thanks

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/AsyncEventListener.java
```diff
@@ -6,13 +6,14 @@
...
+import org.springframework.transaction.event.TransactionPhase;
+import org.springframework.transaction.event.TransactionalEventListener;
 
 @Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@EventListener
+@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
 @Async("eventTaskExecutor")
 public @interface AsyncEventListener
```
fuss86: Name `AsyncEventListener`:
Right now it's now very clear if I use `AsyncEventListener` that it's also related with some `AFTER_COMMIT` transaction state. I mean,  `AsyncEventListener` name may also mean that I can use it just as an async listener without any transactional consideration.

Consider dropping or renaming this class, since `@TransactionalEventListener` may work with other transational phases too.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/Event.java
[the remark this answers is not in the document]
fuss86: I can check tomorrow. Do you remember exception/error message or something ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: 👍

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-raw-layout-provider/src/test/java/org/carlspring/strongbox/config/RawLayoutProviderTestConfig.java
fuss86: I need to say I have the same problem from time to time.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
fuss86: I would try `java.util.concurrent.Executors#newFixedThreadPool(int)`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
carlspring: @sbespalov : Would you like to look into this in a follow-up pull?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-raw-layout-provider/src/test/java/org/carlspring/strongbox/config/RawLayoutProviderTestConfig.java
carlspring: @sbespalov : Thanks for the thorough explanation! We'll have to look into this at some point, as @fuss86 has also had to fix a lot of such context issues in the past. Could either of you add a follow-up task for this in youtrack (with a proper explanation)...? Thanks! :)

sbespalov:
> I can check tomorrow. Do you remember exception/error message or something ?

there was no Exceptions, events just were not processed

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/EventExecutorFactoryBean.java
sbespalov: I have tried separate executor and there was problems with event queue overflow, see my comment above.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
sbespalov: yeah, transactional events already implemented in this PR, I just forget to remove this `Runnable` part, will do it in follow up PR.

strongbox-event-api/src/main/java/org/carlspring/strongbox/event/Event.java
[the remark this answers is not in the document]
sbespalov: there was no Exceptions, events just were not processed

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-raw-layout-provider/src/test/java/org/carlspring/strongbox/config/RawLayoutProviderTestConfig.java
sbespalov: ok

fuss86: SB-1189

PR #760, opened 2018-07-23
SB-1199 Return user authorities when retrieving user data from api

steve-todorov:


strongbox-security/strongbox-user-management/src/main/java/org/carlspring/strongbox/users/service/impl/UserServiceImpl.java
```diff
@@ -79,8 +82,26 @@ public User findByUserName(final String username)
...
+                UserDto userDto = optionalUserDto.get();
+
+                Set<String> authorities = userDto.getRoles()
+                                                 .stream()
+                                                 .map(this::getGrantedAuthorities)
+                                                 .map(this::getAuthoritiesAsString)
+                                                 .flatMap(Collection::stream)
+                                                 .collect(Collectors.toCollection(HashSet::new));
+
+                if(authorities.size() > 0) {
```
carlspring: Braces on new line.

PR #764, opened 2018-07-24
SB-1202 invalid role name fix

fuss86:


strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/users/UserController.java
```diff
@@ -150,7 +166,14 @@ public ResponseEntity getUser(@ApiParam(value = "The name of the user", required
...
     public ResponseEntity getAssignableRoles()
     {
-        return ResponseEntity.ok(new AssignableRoleResponseEntity(this.authorizationConfigService.get().getRoles()));
+        ImmutableSet.Builder builder = ImmutableSet.builder()
+                                                   .addAll(authorizationConfigService.get()
+                                                                                     .getRoles())
+                                                   .addAll(Arrays.stream(Roles.values())
+                                                                 .map(r -> new RoleDto(r.name(), r.getDescription()))
+                                                                 .map(Role::new)
+                                                                 .collect(Collectors.toSet()));
```
steve-todorov: This looks redundant to me because the `assignableRoles` should be a a list of all possible roles available in Strongbox. (That list is used in the UI so that a role can be assigned to a user). However, it looks like you're getting this from `Roles.values()`? 

Also, I was left under the impression that calling `authorizationConfigService.get().getRoles()` should return all of the roles? Can we move this there or maybe as `getAssignableRoles` method?

fuss86: yes, `AuthorizationConfigService` is related with `strongbox-authorization.xml`

PR #771, opened 2018-07-29
SB-1206 authentication suppliers order and password removal from JWT …

fuss86:
…token


steve-todorov:
Thanks for the fix! :)

PR #778, opened 2018-08-03
SB-1217 LDAP Controller updates [ready]

fuss86:


strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
```diff
@@ -0,0 +1,105 @@
...
+            beanFactory.autowireBean(authoritiesMapper);
+            authoritiesMapper.setRolesMapping(form.getConfiguration().getRolesMapping());
+            provider.setAuthoritiesMapper(authoritiesMapper);
+        }
+
+        try
+        {
+            Authentication authentication = provider.authenticate(
+                    new UsernamePasswordAuthenticationToken(form.getUsername(), form.getPassword()));
+            return authentication != null;
```
sbespalov: @fuss86 shouldn't it also check `Authentication.isAuthenticated()` ?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.security.ldap.userdetails.DefaultLdapAuthoritiesPopulator;
+import org.springframework.security.ldap.userdetails.LdapAuthoritiesPopulator;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+@Component
+public class SpringSecurityLdapInternalsTester
```
sbespalov: @fuss86 @steve-todorov how this assumed to be work according our custom authentication implementation? I mean that we assume to have LDAP configuration inside `strongbox-authentication-providers.xml`, so why we need this additional configuration of `LdapAuthenticationProvider` here?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/LdapAuthenticatorConfigurationController.java
```diff
@@ -64,8 +184,8 @@ public ResponseEntity getRolesMapping(@RequestHeader(HttpHeaders.ACCEPT) String
                             @ApiResponse(code = 400, message = "LDAP is not enabled or LDAP role mapping already exists for given LDAP role"),
                             @ApiResponse(code = 500, message = "Failed to add LDAP role mapping") })
     @PostMapping(value = "/rolesMapping/{externalRole}/{internalRole}",
-                 produces = { MediaType.TEXT_PLAIN_VALUE,
-                              MediaType.APPLICATION_JSON_VALUE })
+            produces = { MediaType.TEXT_PLAIN_VALUE,
+                         MediaType.APPLICATION_JSON_VALUE })
     public ResponseEntity addRoleMapping(@PathVariable String externalRole,
```
sbespalov: @fuss86 our role mappings stored only in memory? I mean that we have `AuthoritiesExternalToInternalMapper.rolesMapping` which is seems not persisted.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
fuss86: see SB-1217 point 3. This is to test the configuration.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/LdapAuthenticatorConfigurationController.java
fuss86: This Controller's method updates `AuthoritiesExternalToInternalMapper`  so I'm not sure if I understand the problem, @sbespalov

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
steve-todorov: Yeah, there needs to be an option to test out the ldap configuration settings you've configured from the front-end before you actually apply them. This is necessary so you don't get locked out because of bad/invalid configuration settings (i.e. bad lookup `dn`, wrong group lookup `dn`, `base dn`, etc). Of course if you're manually editing the `strongbox-authentication-providers.xml` you would be skipping the "test" part as it is assumed you know what you're doing.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/LdapAuthenticatorConfigurationController.java
steve-todorov: Cosmetic: I think this class should actually be named `ExternalToInternalAuthoritiesMapper`?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
fuss86: Thanks 👍

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/LdapAuthenticatorConfigurationController.java
fuss86: @carlspring do you confirn ?:)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
sbespalov: maybe I am missing something but as I see if you will make LDAP configuration changes using `LdapAuthenticatorConfigurationController` then all this changes will be lost on strongbox restart and we will get initial configuration from `strongbox-authentication-providers.xml`.

steve-todorov: Hmm.. that shouldn't be working like that. If you change the configuration through the `LdapAuthenticatorConfigurationController` it should actually save those settings in the `strongbox-authentication-providers.xml` and reload/refresh the cached `xml` file (I presume once the `xml` file is loaded it's cached or stored in-memory?). @fuss86 can you confirm?

fuss86: Good catch @sbespalov , we need to update `strongbox-authentication-providers.xml` too. I'll do it in this PR. Thanks for your detective skills :)

>  (I presume once the xml file is loaded it's cached or stored in-memory?). @fuss86 can you confirm?

Yes :) I confirm

sbespalov: >  we need to update strongbox-authentication-providers.xml too

I think it will be better to store LDAP configuration as our other configurations in `strpngbox.xml` and just use it in Spring managed LDAP components.

strongbox-web-core/src/main/java/org/carlspring/strongbox/forms/configuration/security/ldap/LdapConfigurationTestForm.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.validation.constraints.NotNull;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+public class LdapConfigurationTestForm
+{
+
+    @NotEmpty
+    private String username;
```
sbespalov: this probably should be renamed to `testUsername` and `testPassword` for not to confuse people that this is just to test how LDAP authentication works

sbespalov: there is `@JsonProperty` annotation which can map this field to any UI capable JSON property name.

steve-todorov: In that case, if you guys agree that's a more clear property naming - it's fine by me :+1: :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/configuration/security/ldap/support/SpringSecurityLdapInternalsTester.java
fuss86: yeah, for sure I'm not going to reproduce `strongbox-authentication-providers.xml `

fuss86:
moved to #785

PR #782, opened 2018-08-04
SB-1192: Evaluate migrating to Spring Boot

carlspring:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/LayoutProviderRegistry.java
```diff
@@ -31,14 +34,18 @@
...
     public LayoutProviderRegistry()
     {
     }
-    
+
     @Override
     @PostConstruct
     public void initialize()
     {
+        layoutProviders.stream().forEach(lp -> addProvider(lp.getAlias(), lp));
```
carlspring: Uhm... Why are we doing it this way?
I thought that every layour provider was getting itself registered in the registy from within each of the provider's `register` method...?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/LayoutProviderRegistry.java
```diff
@@ -31,14 +34,18 @@
     @Inject
     private ConfigurationManagementService configurationManagementService;
 
+    @Inject
+    private List<LayoutProvider> layoutProviders;
```
carlspring: Where is this being injected from? Do you mean to say that this is in one of the configuration classes? I don't think I like this idea a lot...

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/LayoutProviderRegistry.java
fuss86: In `StrongboxBooter` you use `LayoutProviderRegistry` to initialize repositories. How are you sure that all layout providers have already been registered when we want to use `LayoutProviderRegistry` ? This situation happen in your branch in assembly. Reposiitories couldn't have been initialized properly because `LayoutProviderRegistry` was empty. Layout providers have been registered later. If you don't like this solution, you need to do it in some other way.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/LayoutProviderRegistry.java
fuss86: No. Spring lookups for all known `LayoutProvider`s in the application context and injects them here.

carlspring:
@fuss86 ,

Thanks for your help with pushing this forward!

Here's a little something from the logs that leads me to believe the functionality for scanning for custom authenticators is now broken (in the scope of the Spring Boot migration) : 

```
23:37:35.967 26-09-2018 | INFO  | main                 | o.c.s.resource.ConfigurationResourceResolver       | Using classpath resource path [webapp/WEB-INF/lib]
23:37:35.968 26-09-2018 | DEBUG | main                 | o.c.s.a.r.support.ExternalAuthenticatorsHelper     | class path resource [webapp/WEB-INF/lib] cannot be resolved to URL because it does not exist
23:37:35.968 26-09-2018 | DEBUG | main                 | o.c.s.a.registry.support.AuthenticatorsScanner     | Reloading authenticators registry ...
23:37:35.970 26-09-2018 | INFO  | main                 | o.c.s.resource.ConfigurationResourceResolver       | Try to fetch configuration resource path [/java/opensource/carlspring/strongbox/strongbox-web-core/target/strongbox/etc/conf/strongbox-authentication-providers.xml]
23:37:36.225 26-09-2018 | DEBUG | main                 | o.c.s.a.registry.support.AuthenticatorsScanner     | Scanned authenticators: [org.carlspring.strongbox.authentication.api.impl.xml.PasswordAuthenticator, org.carlspring.strongbox.authentication.api.impl.xml.JwtAuthenticator, org.carlspring.strongbox.authentication.api.impl.xml.SecurityTokenAuthenticator]
```

PR #783, opened 2018-08-06
strongbox-web-integration-tests hotfix; new web-forms module

fuss86:


strongbox-web-forms/src/main/java/org/carlspring/strongbox/forms/configuration/MavenRepositoryConfigurationForm.java
```diff
@@ -1,13 +1,11 @@
...
 
-import org.carlspring.strongbox.providers.layout.Maven2LayoutProvider;
-
 import com.fasterxml.jackson.annotation.JsonTypeName;
 
 /**
  * @author Przemyslaw Fusik
  */
-@JsonTypeName(Maven2LayoutProvider.ALIAS)
+@JsonTypeName("Maven 2")
```
carlspring: Why not use the constant...?

fuss86: because we will have to be dependent on every layout provider. Should it be ?

fuss86:
Merged by courtesy of @steve-todorov

PR #788, opened 2018-09-10
SB-1233 Refactor AccessModel

fuss86:


strongbox-security/strongbox-user-management/src/main/resources/etc/conf/strongbox-security-users.xml
```diff
@@ -62,16 +62,43 @@
...
+                                            <name>ARTIFACTS_VIEW</name>
+                                        </privilege>
+                                        <privilege>
+                                            <name>ARTIFACTS_DEPLOY</name>
+                                        </privilege>
+                                        <privilege>
+                                            <name>ARTIFACTS_DELETE</name>
+                                        </privilege>
+                                        <privilege>
+                                            <name>ARTIFACTS_COPY</name>
```
carlspring: I think we should drop this privilege.

fuss86: `org.carlspring.strongbox.controllers.layout.maven.MavenArtifactController#copy` is protected by the privilege

strongbox-security/strongbox-user-management/src/main/java/org/carlspring/strongbox/users/domain/AccessModelPathPrivileges.java
```diff
@@ -0,0 +1,57 @@
...
+ * @author Przemyslaw Fusik
+ */
+@Immutable
+public class AccessModelPathPrivileges
+        implements Serializable
+{
+
+    private final String path;
+
+    private final boolean wildcard;
```
sbespalov: @fuss86 @steve-todorov what means `widlcard` property?

fuss86: if `true` then accessed privileges are valid in the provided `path` and below the `path` tree (something like `/path/*`).
if `true` then accessed privileges are valid in the provided `path` without subtree

steve-todorov: @sbespalov `wildcard` is used in the UI to "tell" the backend to append `**` to the end of the `path` so that the path becomes `wildcard` and includes all subpaths.

sbespalov: as I see backend uses same `wildcard` attribute without `**`
anyway, I think it will be good to have some javadocs for this to make it clearer for people

steve-todorov:
Thanks for your work on this! :)

PR #855, opened 2018-10-08
Issue-840: Create an RPM distribution

### LINKED ISSUE #840: Create an RPM distribution
We are currently producing `zip` and `gzip` packages for our distribution. We need to be able to create RPM distributions using the [`rpm-maven-plugin`](https://www.mojohaus.org/rpm-maven-plugin/) (or similar). The place where we produce our assemblies is the [`strongbox-distribution`](https://github.com/strongbox/strongbox/tree/master/strongbox-distribution) module.

This task is similar to #841 and depends on #801 to be finished.

v3rtumnus:
Fixes #840


strongbox-distribution/pom.xml
```diff
@@ -355,6 +355,35 @@
                     <tarLongFileMode>gnu</tarLongFileMode>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>rpm-maven-plugin</artifactId>
+                <version>2.2.0</version>
```
carlspring: Could you please define the plugin's version under the `<pluginManagement/>` section of the `strongbox-parent/pom.xml` and keep it versionless here? Thanks! :)

v3rtumnus: Should I do the same for the appassembler plugin, since I am anyway on it?

carlspring: Yeah, it would probably be a good idea. Thanks!

v3rtumnus: Moved versions for all three plugins!

v3rtumnus:
Added the creation of the RPM file and also a service file, as well as a postinstall script to create the relevant user and group.

steve-todorov:
`rpm` and `dpkg-dev` have been added to our alpine images :+1:

carlspring:
In this line of thought, would it be possible to add `apt` and `aptitude` as well for the similar task for creating Debian distributions?

carlspring:
@steve-todorov ,

Could you please try this out and give some feedback?

strongbox-distribution/src/main/resources/postinstall
```diff
@@ -0,0 +1,3 @@
+groupadd strongbox
+useradd -d /usr/local/strongbox -g strongbox strongbox
+chown -R strongbox:strongbox /usr/local/strongbox/
```
carlspring: Please, make sure you add new lines to the end of files when creating, or editing them, as this can sometimes be a problem on different operating systems.

carlspring:
@v3rtumnus , @steve-todorov :

Once this task is done, I think it would be a good idea to go over the [Installation Guide](https://github.com/strongbox/strongbox/wiki/Installation) and update it as necessary.

Things will then need to be tested based on the instructions.

strongbox-distribution/src/main/resources/postinstall
```diff
@@ -0,0 +1,3 @@
+groupadd strongbox
+useradd -d /usr/local/strongbox -g strongbox strongbox
```
steve-todorov: This should not be a regular user but rather a `system` one. Could you please add `--system` here?

steve-todorov:
The produced RPM does not work out of the box in `opensuse leap 15` just yet. 

Strongbox expects to be able to write into `$STRONGBOX_VAULT` which currently is `$STRONGBOX_HOME/../strongbox-vault`. What this mean in this case is `/usr/local/strongbox-vault`. That path does not exist and in most cases on `SELinux` distributions you would not be able to write there to begin with (unless you have manually configured it).

By convention databases are usually stored in `/var/lib/product-name` and I think we should do it the same (thoughts @carlspring?). We should probably create an `/etc/sysconfig/strongbox` file which contains the `STRONGBOX_VAULT` environment variable that points to `/var/lib/strongbox`. 
This would  be the same for Debian, only the path for the file would be `/etc/default/strongbox`.

carlspring:
Closing this in favour of the rebased #1068 .

PR #858, opened 2018-10-10
issue-827 expired artifacts handler

### LINKED ISSUE #827: `maven-metadata.xml` not updated within group+proxy repository
We have a snapshot Maven proxy repository in Strongbox proxying to a Nexus one. The issue is that the information in the `maven-metadata.xml` is not getting updated properly. The snapshot artifact's metadata in Strongbox is [here](https://strongbox.carlspring.org/storages/storage-common-proxies/carlspring/org/carlspring/strongbox/npm-metadata-jackson/1.0-SNAPSHOT/maven-metadata.xml) and the Nexus one is [here](https://dev.carlspring.org/nexus/content/repositories/carlspring/org/carlspring/strongbox/npm-metadata-jackson/1.0-SNAPSHOT/maven-metadata.xml).

The problem is that latest snapshot versions differs:
- In Strongbox, the latest is `1.0-20180917.013403-3`
- In Nexus, the latest is `1.0-20181001.074150-8`

The `maven-metadata.xml` file should be updated at some point with `MavenMetadataManager`, but this is not happening and needs to be fixed.

Steps to reproduce:
* Install Nexus
* Install Strongbox
* Configure Strongbox with a proxy repo to Nexus
* Deploy something to Nexus
* Resolve it from Strongbox
* Deploy a newer version to Nexus
* Resolve the metadata from Strongbox

fuss86:
Fixes #827.


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPath.java
```diff
@@ -306,6 +310,14 @@ public RepositoryPath wrap(Path path)
     {
         return new RepositoryPath(path, fileSystem);
     }
+
+    public boolean isOld()
```
carlspring: This method should probably be called `hasExpired`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPath.java
```diff
@@ -306,6 +310,14 @@ public RepositoryPath wrap(Path path)
...
+            throws IOException
+    {
+        final Instant twentyFourHoursAgo = Instant.now().minus(24, ChronoUnit.HOURS);
```
carlspring: I think a better check would be this: 24 hrs should actually mean "the files expire at midnight". This is, for example, how Maven knows whether to check for a newer version of a timestamped snapshot (as the default update policy is `daily`, meaning: "if there is no cached resource, get it from the remote and if a consecutive request is received later on, don't get the artifact before midnight". I know that some proxy repositories follow this approach as well and I think we should do it too.

This should be configurable via the `strongbox.xml` file. Furthermore, we need to have an option to define the caching period for the metadata (and each layout provider will have to expose a list of path patterns that it considers metadata). I believe this is quite important.

sbespalov: Also this attribute probably should be placed into `RepositoryFileAttributes` so the implementation will be in `AbstractLayoutProvider.getRepositoryFileAttributes`.

carlspring: Also, I believe the expiration interval should be configurable in minutes (for now; later on we might further improve this).

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/layout/MavenMetadataExpiredRepositoryPathHandler.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MavenMetadataExpiredRepositoryPathHandler.class);
+
+    @Inject
+    private RepositoryProviderRegistry repositoryProviderRegistry;
+
+    @Inject
+    private RepositoryPathResolver repositoryPathResolver;
+
+    @Inject
+    private ChecksumCacheManager checksumCacheManager;
```
sbespalov: @fuss86 why you are using `ChecksumCacheManager` here? AFAIK it was only to validate checksum during artifact uploading. Also what was the goal to rework `ChecksumCacheManager`?

fuss86: @sbespalov , please review [this line](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java#L238). It does nothing more than logging error in case of unmatched checksum cases.

I decided to use this cache for real case. I'm validating checksums using this cache.

fuss86:
@carlspring , @sbespalov please review this pull. I'll resolve failing tests tomorrow. Functionality is done.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
```diff
@@ -47,6 +49,20 @@ public static Boolean isArtifact(RepositoryPath path)
         return (Boolean) Files.getAttribute(path, formatAttributes(RepositoryFileAttributeType.ARTIFACT));
     }
 
+    public static boolean wasModifiedAfter(RepositoryPath path,
```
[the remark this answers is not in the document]
fuss86: > this method should be in RepositoryFiles

It's already there

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
```diff
@@ -47,6 +49,20 @@ public static Boolean isArtifact(RepositoryPath path)
...
+                                           Instant timeLinePoint)
+            throws IOException
+    {
+        final FileTime lastModifiedTime = Files.getLastModifiedTime(path);
+        return lastModifiedTime.toInstant().isAfter(timeLinePoint);
+    }
+
+    public static Boolean hasExpired(RepositoryPath path)
```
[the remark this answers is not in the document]
fuss86: > this method should be in RepositoryFiles

and you've commented on `RepositoryFiles` :-)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
fuss86: > also we have ArtifactEntry.lastUpdated which should be used instead of Files.getLastModifiedTime(path) if RepositoryFiles.isArtifact(repositoryPath) == true

it complicates things since we need it in `org.carlspring.strongbox.providers.layout.Maven2LayoutProvider#getRepositoryFileAttributes` to calculate `EXPIRED` `RepositoryFileAttributeType`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
sbespalov: Yeah, sorry :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
sbespalov: I think it should work, can you please check?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFileAttributes.java
```diff
@@ -1,14 +1,24 @@
...
 
 import java.net.URL;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.carlspring.strongbox.artifact.coordinates.ArtifactCoordinates;
+import static org.carlspring.strongbox.providers.io.RepositoryFileAttributeType.ARTIFACT;
+import static org.carlspring.strongbox.providers.io.RepositoryFileAttributeType.ARTIFACT_PATH;
```
carlspring: I think this should be replaced with a wildcard `static import`. There's no point in listing them out one by one, if they're more than 3-4... 5, right...?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
carlspring: Haha! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
```diff
@@ -27,200 +36,97 @@
...
  * @author mtodorov
+ * @author Przemysław Fusik
  */
+@Component
 public class ChecksumCacheManager
+        implements DisposableBean
 {
 
-    private static Logger logger = LoggerFactory.getLogger(ChecksumCacheManager.class);
+    private static final String SHA1_KEY = "sha1";
```
carlspring: Hm... Could you perhaps use something like `EncryptionAlgorithmsEnum.SHA1.getAlgorithm()` here?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
```diff
@@ -27,200 +36,97 @@
...
 
-    /**
-     * Key:     Artifact path
-     * Value:   Artifact checksum.
-     */
-    private Map<String, ArtifactChecksum> cachedChecksums = new LinkedHashMap<>();
+    private static final String MD5_KEY = "md5";
```
carlspring: Hm... Could you perhaps use something like `EncryptionAlgorithmsEnum.MD5.getAlgorithm()` instead?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
```diff
@@ -80,15 +77,22 @@ protected InputStream getInputStreamInternal(RepositoryPath path) throws IOExcep
 
     @Override
     public RepositoryPath fetchPath(RepositoryPath repositoryPath)
-        throws IOException
+            throws IOException
     {
         RepositoryPath result = resolvePathDirectlyFromGroupPathIfPossible(repositoryPath);
-        if (result != null)
+        if (result == null || RepositoryFiles.hasExpired(result))
```
sbespalov: I think it will be better to have it like it was 
```
if ( result != null && !RepositoryFiles.hasExpired(result))
```
this will have less complexity, with no nested `if`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
```diff
@@ -67,14 +49,6 @@ public RepositoryPath fetchRemoteResource(RepositoryPath repositoryPath)
         throws IOException
     {
         Repository repository = repositoryPath.getFileSystem().getRepository();
-        logger.debug(String.format("Checking in [%s]...", repositoryPath));
-
-        final RepositoryPath candidate = preProxyRepositoryAccessAttempt(repositoryPath);
```
sbespalov: I don't think we should remove this logic, we should check expiration in `preProxyRepositoryAccessAttempt` instead.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
```diff
@@ -91,11 +65,6 @@ public RepositoryPath fetchRemoteResource(RepositoryPath repositoryPath)
         
         try (InputStream is = new BufferedInputStream(new ProxyRepositoryInputStream(client, repositoryPath)))
         {
-            if (RepositoryFiles.artifactExists(repositoryPath))
```
sbespalov: same as above

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -211,90 +208,36 @@ private long writeArtifact(RepositoryPath repositoryPath,
...
         if (Boolean.FALSE.equals(checksumAttribute) && !digestMap.isEmpty())
         {
             // Store artifact digests in cache if we have them.
-            addChecksumsToCacheManager(digestMap, repositoryPathId);
+            addChecksumsToCacheManager(digestMap, repositoryPath);
         }
 
         if (Boolean.TRUE.equals(checksumAttribute))
         {
-            byte[] checksumValue = ((ByteArrayOutputStream) aos.getCacheOutputStream()).toByteArray();
```
sbespalov: we shouldn't drop this and all above logic related with checksum validation

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
```diff
@@ -27,200 +36,97 @@
...
 
         return checksum;
     }
 
-    public ArtifactChecksum getArtifactChecksum(String artifactBasePath)
-    {
-        return cachedChecksums.get(artifactBasePath);
-    }
-
-    public boolean validateChecksum(String artifactPath,
```
sbespalov: we shouldn't drop/rework this

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManagerTest.java
```diff
@@ -1,150 +0,0 @@
-package org.carlspring.strongbox.storage.checksum;
```
sbespalov: we shouldn't drop this

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
```diff
@@ -1,20 +1,21 @@
...
  * @author Przemyslaw Fusik
  */
 @Component
-public class MavenArtifactUpdatedEventListener
+public class MavenArtifactMetadataUpdatedEventListener
         extends BaseMavenArtifactEventListener
 {
 
-    @AsyncEventListener
+    @EventListener
```
sbespalov: why this was changed to sync event?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MavenMetadataExpiredRepositoryPathHandler.class);
+
+    @Inject
+    private RepositoryProviderRegistry repositoryProviderRegistry;
+
+    @Inject
+    private RepositoryPathResolver repositoryPathResolver;
+
+    @Inject
+    private ChecksumCacheManager checksumCacheManager;
```
sbespalov: Sorry, but like I have told earlier we shouldn't use `ChecksumCacheManager` within this issue because it have another purpose. It was assumed to be used within artifact uploading to check that artifact matched checksum. 
As about your observation about 'DEBUG' message, then there was a task to make configurable strict checksum validation for uploaded artifacts.
Also this `ChecksumCacheManager` was developed by @carlspring, so maybe he can say something more about it.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/layout/Maven2LayoutProvider.java
```diff
@@ -86,7 +95,7 @@ public boolean isArtifactMetadata(RepositoryPath path)
 
     public boolean isMavenMetadata(RepositoryPath path)
     {
-        return path.getFileName().toString().equals("maven-metadata.xml");
+        return MetadataHelper.MAVEN_METADATA_XML.equals(path.getFileName().toString());
```
sbespalov: :+1:

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/layout/Maven2LayoutProvider.java
```diff
@@ -103,21 +112,27 @@ public boolean isMavenMetadata(RepositoryPath path)
...
-                        result.put(attributeType, value);
-                    }
+                    result.put(attributeType, value);
+
+                    break;
+                case EXPIRED:
+                    value = BooleanUtils.isTrue((Boolean) value) || (isMavenMetadata(repositoryPath)
+                                                                     &&
+                                                                     !RepositoryFiles.wasModifiedAfter(repositoryPath,
+                                                                                                       Instant.now()
```
sbespalov: this probably better to be extracted into variable

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/providers/repository/ParallelDownloadRemoteArtifactTest.java
```diff
@@ -56,8 +60,38 @@
     public void setup()
         throws Exception
     {
-        remoteArtifactInputStream = new RemoteArtifactInputStreamStub(jarArtifact);
-        prepareArtifactResolverContext(remoteArtifactInputStream, true);
+        prepareArtifactResolverContext(null, true);
```
sbespalov: why this and `prepareArtifactResolverContext` was changed?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/ProxyRepositoryProvider.java
```diff
@@ -87,14 +93,21 @@ public RepositoryPath resolvePathForceFetch(RepositoryPath repositoryPath)
             throw e;
         }
     }
-    
+
+    private RepositoryPath handleRepositoryPathExpiration(final RepositoryPath repositoryPath)
+    {
+        final LayoutProvider layoutProvider = layoutProviderRegistry.getProvider(
+                repositoryPath.getRepository().getLayout());
+        return layoutProvider.handleRepositoryPathExpiration(repositoryPath);
```
sbespalov: I think this probably better to be done with sync event instead of explicit new method in `LayoutProvider`.
You can have a look at `ProxyRepositoryProvider.search` method which handle similar case with search against metadata in remote repositories.

sbespalov:
@fuss86 can you please also make a test for group and proxy repositories, which will check that `maven-metadata.xml` expired and updated?
It can probably be something similar to mocked `RestArtifactResolver` that we have in `TestCaseWithMavenArtifactGenerationAndIndexing`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFileAttributes.java
fuss86: I'm using `class count to use import with '*':` equal to `999` (which is close to "never use wildcard" :) ). What's the expected value ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
fuss86: done

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
fuss86: Method `ProxyRepositoryArtifactResolver#fetchRemoteResource` should force fetching artifact from remote as its name states.

This method is called from 

- `ProxyRepositoryProvider#resolvePathForceFetch` (method name indicates `force` indeed)

 and 

- `MavenArtifactFetchedFromRemoteEventListener#downloadArtifactMetadataAtArtifactIdLevelFromRemote` (the method name is also meaningful)

If you want to resolve path (and check if the path exists locally in the meantime), you should use `ProxyRepositoryProvider#fetchPath`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
fuss86: yes, please read my above comment

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManagerTest.java
fuss86: This test is `@Ignore`ed in master.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
fuss86: This listener solves the `II. GroupRepository case` from https://github.com/strongbox/strongbox/issues/827#issuecomment-429534354

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/providers/repository/ParallelDownloadRemoteArtifactTest.java
fuss86: `ParallelDownloadRemoteArtifactTest` overrides `RetryDownloadArtifactTestBase.prepareArtifactResolverContext` 

`ParallelDownloadRemoteArtifactTest#testConcurrentDownload` works concurrently over the same `InputStream` (first argument of original `prepareArtifactResolverContext`). There is a `try-with-resources Statement` in the `ProxyRepositoryArtifactResolver#fetchRemoteResource` (`try (InputStream is = new BufferedInputStream(new ProxyRepositoryInputStream(client, repositoryPath)))`. Threads were closing underlying  `InputStream` so some next Thread faced `Stream was closed` Exception. Overriden method replaces the same instance of mocked `InputStream` by new instance per each Thread.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: We will discuss it with @carlspring soon I hope

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
fuss86: Needs discussion with @carlspring

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
fuss86: FYI: It will force the group `maven-metadata.xml` to be updated if the proxied  `maven-metadata.xml` was re-fetched.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
fuss86: I think this "new coat" of `ChecksumCacheManager` may be useful for both purposes. Besides that, new implementation follows the approach to work on `RepositoryPath`  (and `URI`) instead of `String`s. This PR bases on `ChecksumCacheManager`.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/layout/Maven2LayoutProvider.java
fuss86: Yeah, thanks

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/ProxyRepositoryProvider.java
fuss86: I like this idea, thanks 👍

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFileAttributes.java
carlspring: Hahaha! :)
I think that anything above five static imports from the same class is probably too much... What do you reckon?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/config/StorageApiConfig.java
```diff
@@ -33,16 +32,6 @@
     @Inject
     private List<ArtifactCoordinatesValidator> versionValidators;
 
-    @Bean(name = "checksumCacheManager")
-    ChecksumCacheManager checksumCacheManager()
```
carlspring: Has this been moved elsewhere, or are these settings the defaults now?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
carlspring: Yeah, why has this been dropped? Has it been replaced by something else, or...? What's the reasoning?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/config/StorageApiConfig.java
fuss86: `ChecksumCacheManager` is marked as `@Component` now

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
fuss86: I'll try tomorrow

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: It has been replaced by 

`String checksumValue = ((ByteArrayOutputStream) aos.getCacheOutputStream()).toString(StandardCharsets.UTF_8.name());`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
sbespalov: if you think that `ProxyRepositoryProvider#resolvePathForceFetch` then let's think about better name. But I think we shouldn't change related logic to suit your expectations regarding method name. :)
this and below logic needed to prevent race conditions within concurrent threads, there was an issue about this and currently it works fine. So we really shouldn't drop this

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManagerTest.java
sbespalov: maybe, but we should fix it then instead of dropping

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
sbespalov: > FYI: It will force the group maven-metadata.xml to be updated if the proxied maven-metadata.xml was re-fetched.

yes, but why it should be sync?

the goal for async was the following:
- in case of sync it affects a lot on our benchmark builds
- with async it will not affect on current request or build 

maybe I am missing something but there is no reason to make it sync within this PR

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
sbespalov: > Besides that, new implementation follows the approach to work on RepositoryPath (and URI) instead of Strings
this is very good, and we should keep this `RepositoryPath` improvement, but can you please clarify why you need `ChecksumCacheManager` within `maven-metadata.xml` invalidation?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/providers/repository/ParallelDownloadRemoteArtifactTest.java
sbespalov: >  Threads were closing underlying InputStream so some next Thread faced Stream was closed Exception.

yes, and this was actually a goal of this test to check that artifact was fetched only once within multiple threads.

sbespalov: so the behavior that you describe is correct within this test and assumed to be so.

sbespalov: so we shouldn't change it :)

fuss86: In the concurrent live scenario the won't be the same `InputStream`. There are no exceptions within test.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
fuss86: To cache `maven-metadata.xml` checksum

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/proxied/ProxyRepositoryArtifactResolver.java
sbespalov: let's try to clarify it from another side
there was an issue [SB-1133](https://youtrack.carlspring.org/issue/SB-1133)
so you want to say that your changes will not get it back?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
sbespalov: > It will update group maven-metadata.xml right away if proxied maven-metadata.xml was updated.
right

> We need up-to-date `maven-metadata.xml` on the request time.
how this related with previous statement?

within proxied `maven-metadata.xml` request we will get proxied `maven-metadata.xml`, and this event will update group `maven-metadata.xml`

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
sbespalov: why you need to cache it?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
fuss86: We need up-to-date `maven-metadata.xml` **in the group repository** on the request time.

`maven-metadata.xml` is stored physically in the `group repository`

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
fuss86: Some tools may not send `.sha1` or `.md5` next to the `maven-metadata.xml`. New `ChecksumCacheManager` stores checksums calculated in runtime (using our streams) as well as `.sha1` or `.md5` values if provided.

Without it algorithm described in #827 will be much more inefficient if we no't have checksums (we will be always re-fetching `maven-metadata.xml`).

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/providers/repository/ParallelDownloadRemoteArtifactTest.java
sbespalov: same `InputStream` let us to check the case that only one thread will get valid stream, and other threads should get artifact from hosted cache.

fuss86: We need a way to force fetch artifact from remote. Method mentioned above is the right one. I'll pay attention to your comments in the evening.

sbespalov: like I already commented before you probably need another test to check force fetching within invalidation, this test have another purpose

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/event/artifact/MavenArtifactMetadataUpdatedEventListener.java
sbespalov: ok, lets get back a little and look at the `GroupRepositoryProvider` code below:
```
    public RepositoryPath fetchPath(RepositoryPath repositoryPath)
        throws IOException
    {
        RepositoryPath result = resolvePathDirectlyFromGroupPathIfPossible(repositoryPath);
        if (result != null && !RepositoryFiles.hasExpired(result))
        {
            return result;
        }
        
        return resolvePathTraversal(repositoryPath);
    }
```
we can see here that this event can be triggered only when group `maven-metadata.xml` has expired or not found, and in this case it will return first match `maven-metadata.xml` from group member. So if group member was proxy then `MavenArtifactMetadataUpdatedEventListener` will be triggered, but despite this we will return re-fetched  proxy `maven-metadata.xml` and sync event will not resolve anything here, it will only affect on request time.

Another thing is that this implementation for group repositories seems wrong, because we should always return aggregated `maven-metadata.xml` for group repositories instead of returning first match from group members if there is no aggregated one.

fuss86: Where did you take mentioned code snippet from ? 

It's not from my branch https://github.com/strongbox/strongbox/blob/issue-827/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java#L79 so we shouldn't considering it I guess.

fuss86: > Another thing is that this implementation for group repositories seems wrong, because we should always return aggregated maven-metadata.xml for group repositories instead of returning first match from group members if there is no aggregated one.

@carlspring what do you think ? It's somehow different approach than what we agreed in #827. We can do that as well.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
fuss86: This solution would be fine for `maven-metadata.xml` and maven in general. I think we should consider more generic approach (like cache or database). I choosed the cache for the time being.

carlspring: I believe that if the `maven-metadata.xml` file has not yet been aggregated in a group, attempt to download it from all of the repositories in the group and create a merged representation. This should also be done each time the metadata expires.

sbespalov: > I believe that if the maven-metadata.xml file has not yet been aggregated in a group, attempt to download it from all of the repositories in the group and create a merged representation. This should also be done each time the metadata expires.

also, this will probably better to do in parallel so we have no performance issues

sbespalov: > This solution would be fine for maven-metadata.xml and maven in general. I think we should consider more generic approach (like cache or database). I choosed the cache for the time being.

why do you think so? as for me it will be generic enough.

Anyway I am not sure that your approach will work properly.

If you want to do it this way then please think how you can get checksum from `InputStream` without using `ChecksumCacheManager`.

@carlspring we probably should rename `ChecksumCacheManager` to make it clear that it is relevant only for upload checksum validation.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/config/StorageApiConfig.java
carlspring: I see, thanks! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryFiles.java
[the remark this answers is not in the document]
fuss86: `RepositoryFiles.isArtifact(repositoryPath)` returns `false` for `maven-metadata.xml`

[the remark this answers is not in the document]
fuss86: `repositoryPath.artifactEntry` is `null`

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
fuss86: @carlspring do you have some proposal to rename currently existing [ChecksumCacheManager ](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java)? I'll probably roll it back. Also, I'd like to gain from my implementation too as it's a real cache (please let me know if you disagree).

carlspring: I hadn't thought of this. I think the name is okay, despite the fact that it's only used during uploads... @fuss86 , if we can benefit from your changes, we should!

fuss86:
@carlspring , @sbespalov please review this PR .

Couple of notes:
* Old `ChecksumCacheManager` has been renamed to `CalculatedChecksumCacheManager` (feel free to propose better name)
* I've introduced my own `ChecksumCacheManager` which I benefit from 
* I haven't done anything else in the previous  `ChecksumCacheManager` other than renaiming it (the same goes to the test class)
* I've added 3 test cases: `MavenMetadataExpirationMultipleGroupCaseTest`, `MavenMetadataExpirationProxyCaseTest`, `MavenMetadataExpirationSingleGroupCaseTest`
* I've rolled back all previous changes related with `ParallelDownloadRemoteArtifactTest` 
* Group repositories expired artifacts work as agreed (aggregation)

strongbox-parent/pom.xml
```diff
@@ -69,7 +69,8 @@
         <version.jetty>9.4.6.v20170531</version.jetty>
         <version.lucene>4.8.1</version.lucene>
         <version.logback>1.2.3</version.logback>
-        <version.maven>3.3.9</version.maven>
+        <version.maven>3.6.0</version.maven>
+        <version.maven.strongbox.fork>3.6.1-PR-2-SNAPSHOT</version.maven.strongbox.fork>
```
carlspring: So, are we forking Maven for it's metadata API, or not? (Is this is a leftover...?).

fuss86: yes, see https://github.com/strongbox/maven/pull/2

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
```diff
@@ -106,57 +115,96 @@ protected RepositoryPath resolvePathTraversal(RepositoryPath repositoryPath) thr
...
+        return result;
+    }
+
+    private RepositoryPath invokeResolvePathActions(List<Callable<RepositoryPath>> resolvePathActions)
+    {
+        List<Future<RepositoryPath>> answers;
+        try
+        {
+            // TODO invokeAll timeout interrupt ?
+            answers = executorService.invokeAll(resolvePathActions);
```
sbespalov: I think we can use `Collection.parallelStream()` here instead of custom executor.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            return false;
+        }
+        if (!MetadataHelper.MAVEN_METADATA_XML.equals(repositoryPath.getFileName().toString()))
+        {
+            return false;
+        }
+
+        Repository repository = repositoryPath.getRepository();
+        RepositoryProvider provider = repositoryProviderRegistry.getProvider(repository.getType());
+        return provider instanceof ProxyRepositoryProvider;
```
sbespalov: I think we can check it other way instead:
```
repository.getRemoteRepository() != null
```

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
```diff
@@ -1,226 +1,114 @@
...
- * - a checksum is not claimed within cachedChecksumLifetime, trigger an event and log
- * this, then remove the checksum from the cache. (If this checksums set is empty,
- * remove the respective Checksum from the cachedChecksums).
- *
- * @author mtodorov
+ * @author Przemysław Fusik
  */
+@Component
 public class ChecksumCacheManager
+        implements DisposableBean
```
sbespalov: @fuss86 how you assume to invalidate this cache?

and also if you will invalidate it somehow then it will lose checksums that you need.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: Then I'd have to use custom `ForkJoinPool` because I wouldn't want to use the default one, because by default it is configured to use as many threads as there are cores in the system.

What's the benefit of choosing `Collection.parallelStream()` ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
fuss86: There is no cache invalidation right now. We can think about it and prepare a follow-up task later for this , but right now it works this way: once the new cache value has been calculated - it replaces old one in cache.

fuss86:
@carlspring , @sbespalov once again request for your review here ;-)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
sbespalov: > What's the benefit of choosing Collection.parallelStream() ?

As I understand `ForkJoinPool.commonPool()`  use same parallelism as you calculate manually for `@CommonExecutorService` (number of cores minus one), but it also use caller thread as worker thread. 
So I think `@CommonExecutorService` redundant and less productive then default `ForkJoinPool`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
fuss86: There won't be OOM because there is [default config setup for this cache](https://github.com/strongbox/strongbox/blob/issue-827/strongbox-data-service/src/main/java/org/carlspring/strongbox/config/HazelcastConfiguration.java#L45) so entries will be evicted when free heap size is below 1000 MB. Besides that checksums are cheap and cache is the fastest solution. 

However, I'm keen on re-implementing it now if you can provide some better alternative (having the code base we already have).

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: > As I understand ForkJoinPool.commonPool() use same parallelism as you calculate manually for @CommonExecutorService (number of cores minus one), 

Looks like you're right, thanks for this info :-)

> but it also use caller thread as worker thread.

Where did you find it ?

> So I think @CommonExecutorService redundant

Ok, I'm convinced now :-)

>  and less productive then default ForkJoinPool.

Hmm.. why so ? :-)

sbespalov: you can check it with this code:
```
System.out.println(String.format("This is the caller thread: [%s]", Thread.currentThread().getName()));
        IntStream.range(0, Runtime.getRuntime().availableProcessors())
                 .parallel()
                 .forEach(i -> {
                     try
                     {
                         Thread.sleep(100);
                     }
                     catch (InterruptedException e)
                     {
                         e.printStackTrace();
                     }
                     System.out.println(Thread.currentThread().getName());
                 });
```

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
sbespalov: I suggested the solution previously, you can directly store calculated checksums within artifact upload (instead of cache that you currenly use) and then overwrite it if original checksum uploaded.

carlspring: My initial implementation was such that when an artifact would get deployed:
* It's checksum would get added to the cache, based on the calculations of the `MultipleDigestInputStream`.
* Items in the cache would expire in a certain amount of time.
* The cache, (albeit quite primitive in my implementation), would remove entries, if they weren't accessed after a certain time (if I recall correctly).
* When an artifact's checksums were uploaded, they would get compared to the ones in the cache and:
  * If they matched, the checksum would get removed from the cache
  * If they didn't match, a warning would be printed out and _**then**_ checksum would be removed from the cache.

Obviously, this was a very primitive implementation, as it neither used EhCache, nor Hazelcast. And during the various refactorings this behavior may have changed.

@fuss86 : How does it work now?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/config/StorageApiConfig.java
```diff
@@ -33,30 +33,30 @@
...
+        calculatedChecksumCacheManager.setCachedChecksumExpiredCheckInterval(300000);
+        calculatedChecksumCacheManager.setCachedChecksumLifetime(60000);
 
-        return checksumCacheManager;
+        return calculatedChecksumCacheManager;
     }
 
     @Bean(name = "versionValidators")
-    LinkedHashSet<ArtifactCoordinatesValidator> versionValidators()
+    protected LinkedHashSet<ArtifactCoordinatesValidator> versionValidators()
```
carlspring: Out of curiousity, why are these all `protected`?

fuss86: `Codacy` reported a warning to explicitly provide visibility - in my opinion, we could shut it out, is possible, @steve-todorov

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: > because it utilize caller thread as well

Ok, so summarizing, `Collection.parallelStream()` uses #`total number of available cores` threads = (`number of cores minus one` (from `ForkJoinPool.commonPool()`)+ `the caller thread`.

This whole conversation led me to the conclusion that we shouldn't use `common` pools all over the code base in our specific environment. Using the same `ForkJoinPool.commonPool()` (or `@CommonExecutorService` in many places may lead to block other parts of the code (other streams) using the same `common` pool. 

I'm gonna use specific `ExecutorService`/`thread pool` for `GroupRepositoryProvider`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
fuss86: @carlspring I wrote about it above:

```
* Old ChecksumCacheManager has been renamed to CalculatedChecksumCacheManager (feel free to propose better name)
* I've introduced my own ChecksumCacheManager which I benefit from
* I haven't done anything else in the previous ChecksumCacheManager other than renaiming it (the same goes to the test class)
```

I see that it's all not clear so I'll change my implementation.

fuss86: > My initial implementation was such that when an artifact would get deployed:

I'll revert it all back.

fuss86: > I suggested the solution previously, you can directly store calculated checksums within artifact upload (instead of cache that you currenly use) and then overwrite it if original checksum uploaded.

OK

fuss86: > Or you can even just drop the cache, because your logic seems implies case when there is no checksum (which is the same as evicted from cache checksum).

"My" `ChecksumCacheManager` was used to handle expiration of the `RepositoryPath` so thiese checksums were used.

fuss86:
@sbespalov, cache related things are now rolled back

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
sbespalov: > Using the same ForkJoinPool.commonPool() (or @CommonExecutorService in many places may lead to block other parts of the code (other streams) using the same common pool.

with `ForkJoinPool.commonPool()` there is no way to be completely blocked because it will always have at least 1 available thread (which is caller therad) .

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/storage/checksum/ChecksumCacheManager.java
sbespalov: > "My" ChecksumCacheManager was used to handle expiration of the RepositoryPath so these cached checksums were used.

sorry, but I dont understand what you mean here

strongbox-parent/pom.xml
```diff
@@ -69,7 +69,7 @@
         <version.jetty>9.4.6.v20170531</version.jetty>
         <version.lucene>4.8.1</version.lucene>
         <version.logback>1.2.3</version.logback>
-        <version.maven>3.3.9</version.maven>
+        <version.maven>3.6.0</version.maven>
```
carlspring: If we're changing this, then we'll also need to update the [Requirements](https://github.com/strongbox/strongbox#requirements) section of the `README.md` accordingly.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
```diff
@@ -106,57 +116,69 @@ protected RepositoryPath resolvePathTraversal(RepositoryPath repositoryPath) thr
...
+    private void invokeResolvePathActions(List<Callable<RepositoryPath>> resolvePathActions)
+    {
+        try
+        {
+            // TODO invokeAll timeout interrupt ?
+            executorService.invokeAll(resolvePathActions);
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
```
sbespalov: why this needed? 
as I understand `InterruptedException` occurs when thread already been interrupted :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
```diff
@@ -59,43 +52,60 @@
...
     @Inject
     private RepositoryPathResolver repositoryPathResolver;
-    
+
+    private ExecutorService executorService;
+
+    @Override
+    public void afterPropertiesSet()
+    {
+        executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
sbespalov: this will will cause limited group repositories concurrency and request blocking, as you noticed. 
if you don't want to use `Collection.parallelStream()`, like was suggested, then can you please use `EventsConfig.eventTaskExecutor`reference here instead?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
```diff
@@ -169,7 +169,14 @@ protected boolean isChecksum(String fileName)
                 value = repositoryPath.getRepository().getId();
 
                 break;
-                
+            case EXPIRED:
+                value = false;
+
+                break;
+            case REQUIRES_GROUP_AGGREGATION:
```
sbespalov: This seems to me that such attribute is not really good thing for common attributes, because in fact it's only for maven. 
Maybe we should consider new method in `RepositoryProvider` for repository group metadata resolution as was suggested in channel?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -220,6 +218,9 @@ private long writeArtifact(RepositoryPath repositoryPath,
         {
             // Store artifact digests in cache if we have them.
             addChecksumsToCacheManager(digestMap, repositoryPathId);
+            //
+            artifactEventListenerRegistry.dispatchEvent(
```
sbespalov: why we need an event for this? 
I think it can be implemented directly in `ArtifactManagementService`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
sbespalov: to be clear I mean that `Collection.parallelStream()` will not completely block other requests because it use caller thread, so at least one thread will be available.
other option is  `EventsConfig.eventTaskExecutor` which use Servlet Container thread pool, so concurrency will be totally limited with our application performance.
as for me then I would like to try `Collection.parallelStream()` and see how it will work in production.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
sbespalov: ok, after revisiting SB-1309 it seems I see what you mean with `firstMatch` logic in `GroupRepositoryProvider` and this attribute, but this attribute special case for real :)
so I suggest to change it with sync event like we have for expired metadata fetch.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: https://www.ibm.com/developerworks/library/j-jtp05236/index.html

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
fuss86: I'll consider that in the evening

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: It should be flexible for each lauout provider to handle this event. Maven case may not be the same way handled like NPM for example. I suggest to leave it as it is.

sbespalov: I think it will be the same. 

my teacher on the OOP at the institute said a very clever thing: “you don’t need to do that is not required of you”.

This is the only useful thing that I learned from him :smile:

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: > to be clear I mean that Collection.parallelStream() will not completely block other requests because it use caller thread, so at least one thread will be available.

I'm worried about blocking all threads from servlet container pool.

> EventsConfig.eventTaskExecutor which use Servlet Container thread pool, 

The same here

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: > I think it will be the same.

But we are not sure, and event approach is more flexible. 

@carlspring what do you think ? The thing is: we want to handle calculated checksums. In maven case, we will store them in the `.sha1` or `.md5` files. There might be other handlers for other layout providers, do you know some other case maybe ?

carlspring: NuGet uses `sha512`. I think it would be better to be more flexible here.

sbespalov: afaik you already have layout specific checksums in `digestMap`

sbespalov: also, events probably is not the best way to get flexibility here

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/repository/GroupRepositoryProvider.java
fuss86: As you said, `Collection.parallelStream() ` will utilize caller thread too (a thread from servlet container pool). If 10 independent requests arrive to the server then 10 threads from the servlet container pool could be utilized and blocked here.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/config/StorageApiConfig.java
steve-todorov: > Codacy reported a warning to explicitly provide visibility
I disabled it in codacy.. hopefully it's not reporting it anymore :)

strongbox-parent/pom.xml
steve-todorov: Let me know if this is a requirement so I can upgrade our ci docker images and jenkins jobs (where necessary).

fuss86: It should not be necessary. I'll roll it back

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: > also, events probably is not the best way to get flexibility here

On the contrary I believe it really is.

fuss86: > afaik you already have layout specific checksums in digestMap

AFAIS, currently `SHA1` and `MD5` are supported but it can be extended in the future.

I want to leave it as it is.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
fuss86: I think we don't need to make it sync. Here is the most common case:

1. Request for `maven-metadata.xml`
2. Fetch remote checksum of `maven-metadata.xml`
3. Compare remote checksum with local checksum
4. If checksums don't match, re-fetch `maven-metadata.xml`
5. Storing calculated checksum for re-fetched `maven-metadata.xml` may be done in async, since the  `maven-metadata.xml` will expire after a minute (we can assume async event will be called within this time).

If my assumptions are bad, we will probably be notified about it in Jenkins.

strongbox-parent/pom.xml
steve-todorov: Okay, many thanks! :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
sbespalov: please have a look at `LayoutFileSystemProvider.decorateStream` method, it provides layout specific `digestAlgorithmSet` for `LayoutOutputStream` you have.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
sbespalov: yes, but I mean not this :)
I mean that we shouldn't add this extra `REQUIRES_GROUP_AGGREGATION` attribute.
Currently you using it to get ` boolean firstMatch` and then decide that you need to aggregate metadata based on it.
So we should drop `REQUIRES_GROUP_AGGREGATION` and trigger an event just like you have it in `ProxyRepositoryProvider` for expired paths.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
fuss86: ok, now I understand :)

Done

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/layout/AbstractLayoutProvider.java
fuss86: ok, it can be done this way too.

Done.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/main/java/org/carlspring/strongbox/providers/io/MavenMetadataExpiredRepositoryPathHandler.java
```diff
@@ -0,0 +1,127 @@
...
+        return repository.getRemoteRepository() != null;
+    }
+
+    @Override
+    public void handleExpiration(final RepositoryPath repositoryPath)
+            throws IOException
+    {
+        Decision refetchMetadata = determineMetadataRefetch(repositoryPath,
+                                                            EncryptionAlgorithmsEnum.SHA1);
+        if (refetchMetadata == I_DONT_KNOW)
```
carlspring: Hahaha!!! :D

steve-todorov: I think @fuss86 just received +9999 player points for this. :joy:

carlspring:
Thanks for all the hard work and the thorough discussions! :)

PR #882, opened 2018-10-17
SB-1271 Upgrading sonar plugin version

steve-todorov:


steve-todorov:
Failures unrelated to changes.

Issue #538, opened 2018-01-25
Create checkstyle rules for the project based on the coding convention


carlspring:
We have a clearly defined [Coding Convention](https://github.com/strongbox/strongbox/wiki/Coding-Convention) in the wiki. It would be great to have rules created for this, so that the convention could be checked by Jenkins when new pull requests have been made.

leonorader:
Your coding convention looks awesome! :) I would like to work on this!
If you accept, could you please decide if I should add it as a plugin, too, or just make a ruleset, so that you can import it into you IDEs?

carlspring:
Hi,

Thanks for the feedback! We'd love the help!

Could you please share the pros and cons of each of these?

leonorader:
Sure! 

Adding as a maven build step:
- PRO: if you have any Checkstyle violation in the code then you will not be able to finish a build, until all of these are corrected (it can go very well if you really followed all the conventions, but it might cause hundreds, even thousands of errors, so it might take a lot of time to fix)
- CON: taking it seriously will allow you to make sure that all the code is consistent and follows all the rules

Using only as a plugin in your IDE:
- PRO: will be able to build even if there are millions of violations
- CON: you might have non-consistent code

So I would suggest to first make the ruleset, then first try to run a check on the code and if it's all okay, then I will add it as a build step. Otherwise fixing those violations can be started. 

:)

carlspring:
Thank you for your explanation!

I believe what you're suggesting is a good idea. Please, feel fre to proceed with the ruleset and we'll see how it works out. :)

steve-todorov:
This has been fixed [PR-889](https://github.com/strongbox/strongbox/pull/889).

PR #889, opened 2018-10-20
checkstyle - intial version of ruleset added

leonorader:
Issue #538 
So I added the initial version of the checkstyle ruleset. I run in on the code, and it had many warnings, then I run in on the example code on the guildelines page and it was all good.

A few more things to mention:
- it will check only Java files for now
- read more on the checks [here](http://checkstyle.sourceforge.net/checks.html)
- I set the level to warning, but later it can be changed to error if you want
- each warning will mention the name of the check that causes the violation so you can read about it
- I went though the codebase and I haven't found any real javadoc, but it could be also configured later

To be able to run a check, you can add these lines to the main pom:
```
<build>
    <plugins>
        <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-checkstyle-plugin</artifactId>
            <version>3.0.0</version>
            <dependencies>
                <dependency>
                    <groupId>com.puppycrawl.tools</groupId>
                    <artifactId>checkstyle</artifactId>
                    <version>8.13</version>
                </dependency>
            </dependencies>
            <configuration>
                <configLocation>strongbox-checks.xml</configLocation>
                <consoleOutput>true</consoleOutput>
            </configuration>
        </plugin>
    </plugins>
</build>
```

And run checkstyle:check. You will see the list of warnings in the console. :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,5 +101,34 @@
                 </plugins>
             </build>
         </profile>
+		
+		<profile>
+            <id>checkstyle</id>
+
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+
+            <build>
+                <plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-checkstyle-plugin</artifactId>
+						<version>3.0.0</version>
+						<dependencies>
+							<dependency>
+								<groupId>com.puppycrawl.tools</groupId>
+								<artifactId>checkstyle</artifactId>
+								<version>8.13</version>
+							</dependency>
+						</dependencies>
+						<configuration>
+							<configLocation>strongbox-checks.xml</configLocation>
+							<consoleOutput>true</consoleOutput>
+						</configuration>
+					</plugin>
+				</plugins>
+            </build>
+        </profile>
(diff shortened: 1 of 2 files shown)


carlspring:
@leonorader,

I've had a look at the pull and it looks good (although, I haven't yet had a chance to try out myself)! One small request -- would you mind moving the `maven-checkstyle-plugin` into a Maven `<profile>` called `checkstyle`? We wouldn't want to run this upon every build. It would have to run only on our CI server, as it would otherwise slow things down significantly.

leonorader:
sure! would you like to have this profile in the parent pom?

carlspring:
Yeah, I think that would be the best place! Thanks! :)

carlspring:
Hi,

Could you please sign our ICLA?

leonorader:
Of course! :)

Anyway could you please help why Jenkins has test failures? I really don't know what I should do exactly.
2 tests failed in this file: org.carlspring.strongbox.providers.repository.MavenProxyRepositoryProviderTestIT.whenDownloadingArtifactMetadataFileShouldBeMergedWhenExist

carlspring:
Please, disregard these two failures.

PR #915, opened 2018-11-06
SB-1296 resource handler before request mapping handler

fuss86:


steve-todorov:
Thanks for fixing this!

PR #921, opened 2018-11-07
SB-1301 Surefire tests cannot be run in a single module anymore - after migration to Junit5

fuss86:
…ation to Junit5


carlspring:
@ptirador : FYI.

PR #922, opened 2018-11-07
SB-1300: Integration tests cannot be run in a single module anymore - after migration to Junit5

fuss86:


carlspring:
@ptirador : FYI.

PR #924, opened 2018-11-08
SB-948: Configure spring to return a custom 404 error page which actually serves index.html

sbespalov:
additionally following was done:
- jetty resources cleanup
- distribution logging profiles fixed


steve-todorov:
Thanks for the fix! :)

PR #957, opened 2018-12-12
SB-1326 database indexes sync

fuss86:


strongbox-data-service/src/main/java/org/carlspring/strongbox/data/service/SynchronizedCommonCrudService.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.carlspring.strongbox.data.domain.GenericEntity;
+
+import javax.inject.Inject;
+
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
+import com.orientechnologies.orient.core.id.ORID;
+
+public abstract class SynchronizedCommonCrudService<T extends GenericEntity>
```
sbespalov: @fuss86 we already have a synchronization within Business layer, I don't think we should also have it in DAO by default on every save operation.

sbespalov:
@fuss86 as I say we already have a synchronization. 
The most of problems within Unit Tests is that we don't use `ArtifactManagementService` there so it not synchronized in test env unlike application env.
to fix this synchronization in tests we can just make `MavenTestCaseWithArtifactGeneration.generateArtifact` (and similar) methods synchronized.

fuss86:
> The most of problems within Unit Tests is that we don't use `ArtifactManagementService` there so it not synchronized in test env unlike application env.

Yes, but please be aware that there are also other places (excluding `ArtifactManagementService`) where we use DAO layer (e.g. `NpmPackageFeedParser#saveArtifactEntry`, `NugetRepositoryFeatures#parseFeed`, `AsyncArtifactEntryHandler`) and all those places require repository path locking engagement. We can imagine other methods implemented in the future that will use DAO layer too in the similar way.

With additional DAO layer synchronization we gain extra safety level. If there is a gap in the business layer, DAO layer synchronization comes to the rescue.

What's the potential overhead of adding extra synchronization on the DAO layer if we assume business layer does its job well enough ? Time needed to acquire/release one lock, without blocking other threads (since business layer did it). If you compare the full build times of strongbox-against-strongbox of this PR build with master, you will not notice any difference. 

Isn't it worth to try this solution in production ?

fuss86:
Ok, thanks for your insight, @sbespalov . 

@carlspring, I'm leaving the decision up to you. Please read our above comments. Thanks !

fuss86:
@carlspring so it looks like the decision is not easy, so I'll follow @sbespalov `s suggestion

carlspring:
> @carlspring so it looks like the decision is not easy, so I'll follow @sbespalov `s suggestion

I agree with @sbespalov . I've been following this discussion closely which is why I haven't rushed ahead with merging this.

carlspring:
What is the status of this PR?

fuss86:
I think we need to close it and prepare a solution depending on the existing issues where the problem occurs.

PR #959, opened 2018-12-17
SB-1232 db snapshot import [in progress]

fuss86:


fuss86:
Closing in favor to #963

PR #962, opened 2018-12-26
SB-1289: fix windows build

sbespalov:


steve-todorov:
Thanks for fixing this! :)

PR #980, opened 2019-01-14
SB-1350 Increase max allowed chunk length to 64MB - orientDb

fuss86:


steve-todorov:
Thanks for fixing this! :)

PR #981, opened 2019-01-14
SB-1351 strongbox-distribution is not able to start - fix

fuss86:


carlspring:
Thanks for fixing this! :)

PR #987, opened 2019-01-20
SB-1253 artifact group

fuss86:


strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Should be unique per type
+     */
+    private String name;
+
+    private String type;
+
+    public ArtifactGroup()
+    {
+        type = getClass().getName();
```
sbespalov: do we really need this? `OrientDB` have class metadata by default

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
```diff
@@ -85,24 +106,28 @@ private boolean artifactEntryIsSavedForTheFirstTime(ArtifactEntry artifactEntry)
...
+                           setLastVersionTag(lastVersionEntry, entity, lastVersionTag);
+                       }
+              );
+    }
+
+    private <S extends ArtifactEntry> void setLastVersionTag(ArtifactEntry lastVersionEntry,
+                                                             S entity,
+                                                             ArtifactTag lastVersionTag)
+    {
+        Optional<ArtifactCoordinates> lastVersionCoordinates = Optional.ofNullable(
```
sbespalov: `artifactCoordinates` shouldn't be `null`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        params.put("name", name);
+        params.put("type", type.getName());
+
+        String sQuery = buildQuery(params);
+
+        OSQLSynchQuery<Long> oQuery = new OSQLSynchQuery<>(sQuery);
+        oQuery.setLimit(1);
+
+        // make sure there are no duplicates per `params`
+        lock(params);
```
sbespalov: I think we can use `NeedRetryException` here instead of locking.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
fuss86: it's from [master](https://github.com/strongbox/strongbox/blob/master/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java#L105)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
fuss86: Without it, I won't be able to use [this](https://github.com/strongbox/strongbox/blob/master/strongbox-data-service/src/main/java/org/carlspring/strongbox/data/service/CommonCrudService.java#L273) method. I wish there could be only one `ArtifactGroupServiceImpl` (without the need to extend services per group) with single `ArtifactGroupServiceImpl#getEntityClass` returning `ArtifactGroup.class`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
fuss86: yeah, thanks :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
```diff
@@ -47,6 +50,20 @@ protected boolean identifyEntity(AbstractArtifactCoordinates entity)
...
     }
 
+    @Override
+    protected <S extends AbstractArtifactCoordinates> S cascadeEntitySave(AbstractArtifactCoordinates entity)
+    {
+        try
+        {
+            return super.cascadeEntitySave(entity);
+        }
+        catch (OConcurrentModificationException ex)
```
sbespalov: why this needed?
I thought we already fixed all of such issues with `AritfactEntry` and `ArtifactCoordinates`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
sbespalov: seems I see what you mean, it probably [this](https://github.com/strongbox/strongbox/pull/987/files#diff-6af8485da2f1ab69413f1f04f4fdb475R85) place :)
anyway there is not a big deal to make a custom query `String.format("select * from %s where name=:name", artifactGroupClass)`.
but I think that we probably need to make `ArtifactGroup.name` attribute unique, so you can always use just `ArtifactGroup` class when selecting by `name`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
```diff
@@ -42,7 +46,13 @@
 
     @Inject
     private ArtifactTagService artifactTagService;
-    
+
+    @Inject
+    private ArtifactGroupService artifactGroupService;
+
+    @Inject
+    private LayoutProviderRegistry layoutProviderRegistry;
```
sbespalov: it seems not used here.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
```diff
@@ -59,15 +69,27 @@
         ArtifactCoordinates coordinates = entity.getArtifactCoordinates();
         if (coordinates == null)
```
sbespalov: yeah, I see that this is like in `master` but currently we shouldn't have any `AritfactEntry` without `ArtifactCoordinates`.
So can you please drop this `if` statement and check that we have `not null` constraint for `ArtifactEntry.artifactCoordinates`?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
```diff
@@ -51,6 +45,9 @@
 
     private Integer downloadCount = Integer.valueOf(0);
 
+    @Transient
+    private Set<ArtifactGroup> artifactGroups;
```
sbespalov: I found that this `Transient` field really confusing me. Can we drop it?

sbespalov:
@fuss86 can you please share your thoughts about [this](https://youtrack.carlspring.org/issue/SB-1253#focus=streamItem-84-1736-0-0) comment?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
fuss86: > but I think that we probably need to make ArtifactGroup.name attribute unique, 

It is already made unique by `type` and `name`, see [here](https://github.com/strongbox/strongbox-db/pull/5/files#diff-9eb19f9d18f36a41149aae0396429bbaR17).

> so you can always use just ArtifactGroup class when selecting by name.

I think that group name should be unique in a specific group type, but not in general.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
fuss86: yes, thanks

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
fuss86: This way we pass the `artifactGroups` to the service layer and we can save artifact entries related with it (see [here](https://github.com/strongbox/strongbox/pull/987/files#diff-ca2af8f423f4756b1737ad2178c1101dR82)). Otherwise we would have to pass it as an additional parameter to the `ArtifactEntryServiceImpl#save(S, boolean)` method (keeping the `artifactGroups`  transactional with `ArtifactEntry`).

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
fuss86: see [here](https://jenkins.carlspring.org/job/strongbox/job/strongbox/job/PR-987/13/testReport/)

Short explanation: when we save artifact group [here](https://github.com/strongbox/strongbox/pull/987/files#diff-ca2af8f423f4756b1737ad2178c1101dR85) all related `artifactEntries` and `artifactCoordinates` have updated `entityVersion`.

fuss86: This is related with `org.carlspring.strongbox.data.tx.OEntityUnproxyAspect` and `com.orientechnologies.orient.core.annotation.OVersion`

sbespalov: this should not happen because we have synchronization by `ArtifactCoordinates.getId()` [here](https://github.com/strongbox/strongbox/blob/SB-1253/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/providers/io/RepositoryPathLock.java#L66).

can you please double check it?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
sbespalov: I think that we probably need to have `ArtifactGroupService.addArtifactToGroup` method, and move `last-version` tag update logic there.

> keeping the artifactGroups transactional with ArtifactEntry

we have such transaction in `ArtifactManagementService`

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
fuss86: [Here](https://jenkins.carlspring.org/job/strongbox/job/strongbox/view/change-requests/job/PR-987/16/) is the PR build with this catch.

fuss86: And [here](https://jenkins.carlspring.org/job/strongbox/job/strongbox/job/PR-987/17/) is without this catch. I'll try to analyze it deeper.

fuss86: The reason is that when we fetch the `ArtifactGroup`, all related `artifactEntries` are eagerly fetched too (`@OneToMany(fetch = FetchType.LAZY)` does not work in our case in orientDb :( ). Inside these fetched collection of `artifactEntries`  there might be cases with the same `artifactCoordinates` (let say artifactEntry1 and artifactEntry2). If we want to remove the `last-version` tag from  `artifactEntry1` first and then from `artifactEntry2` we've got `OConcurrentModificationException ` because they share the same `artifactCoordinates` .

fuss86: I also tried https://orientdb.com/docs/last/Fetching-Strategies.html without luck. It doesn't work in our case.

fuss86: It can also be solved like [this](https://github.com/strongbox/strongbox/pull/987/commits/8ec7ebbfc69f722fd5df46b7d3f78bf952b18d7d).

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
fuss86: I got rid of this field.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
fuss86: And build is green with it. What do you think, @sbespalov ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactIdGroup.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public String getArtifactId()
+    {
+        return getName();
+    }
+
+    public static String createName(ArtifactCoordinates artifactCoordinates)
+    {
+        return String.format("%s/%s", artifactCoordinates.getClass().getSimpleName(), artifactCoordinates.getId());
```
sbespalov: `artifactCoordinates.getClass()` can be DB proxy, so this can work not as expected

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    private <T extends ArtifactGroup> T instantiateNew(Class<T> type,
+                                                       String name)
+    {
+        T artifactGroup;
+        try
+        {
+            artifactGroup = type.getConstructor(String.class)
```
sbespalov: this is not really generic :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
```diff
@@ -75,7 +75,7 @@ public void setRepositoryId(String repositoryId)
         this.repositoryId = repositoryId;
     }
 
-    public ArtifactCoordinates getArtifactCoordinates()
+    public AbstractArtifactCoordinates getArtifactCoordinates()
```
sbespalov: why we need to change this?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
sbespalov: @fuss86 I suggest to leave concurrency and make `last-version` update work as expected first.
I think we have some confusions about it.

sbespalov:
@fuss86 can you please provide a `last-version` management tests for cases like [this](https://youtrack.carlspring.org/issue/SB-1253#focus=streamItem-84-1774-0-0).

fuss86:
closed accidentally

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
fuss86: yes, it is. Every artifact group has a name. I would make it the only one constructor without orientDb needs to have no-args constructor too.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactIdGroup.java
fuss86: We unproxy everything from the orientdb so it wouldn't happen in orientdb case. However, I agree there is some potential bug field here.

What do you think about adding and using here something like this:
```
.....
public interface ArtifactCoordinates  ....
{
     ....
    Class<T> getType();

}
```

and in maven case for example:
```
public class MavenArtifactCoordinates ....
{
    @Override
    public Class<MavenArtifactCoordinates> getType()
    {
        return MavenArtifactCoordinates.class;
    }
```
for security needs ?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
fuss86: To get objectId [here](https://github.com/strongbox/strongbox/pull/987/commits/8ec7ebbfc69f722fd5df46b7d3f78bf952b18d7d#diff-ca2af8f423f4756b1737ad2178c1101dR158). `artifactCoordinates` is indeed `AbstractArtifactCoordinates ` field.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
sbespalov: the thing is that you can't say what `ArtifactGroup` implementations will be in the future. And there can be not only `name` needed for some implementations.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactIdGroup.java
sbespalov: We already have `@ArtifactCoordinatesLayout` which defines layout name for every `ArtifactCoordinates` implementation.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
fuss86: No longer valid.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactIdGroup.java
fuss86: Changed to per repository solution.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactEntry.java
fuss86: rolled back

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
```diff
@@ -81,52 +103,77 @@ private boolean artifactEntryIsSavedForTheFirstTime(ArtifactEntry artifactEntry)
...
+        Set<ArtifactEntry> lastVersionEntries = findLastVersionArtifactEntries(artifactGroup,
+                                                                               lastVersionTag,
+                                                                               entity);
+        lastVersionEntries.stream()
+                          .forEach(lastVersionEntry -> {
+                              setLastVersionTag(lastVersionEntry, entity, lastVersionTag);
+                          });
+    }
+
+    private <S extends ArtifactEntry> void setLastVersionTag(S lastVersionEntry,
```
sbespalov: maybe we should call this method `checkAndUpdateLastVersionTagIfNeeded`? :)
I mean that probably there should be clearer name for this method, @fuss86  what do you think?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+            optional = tryFind(type, properties);
+            if (optional.isPresent())
+            {
+                return optional.get();
+            }
+            throw ex;
+        }
+    }
+
+    private <T extends ArtifactGroup> T instantiateNew(Class<T> type,
```
sbespalov: @fuss86 sorry but I don't really like this way to create `ArtifactGorup` instance.
It's seems to me that we should better have concert `ArtifactGrouptService` implementations, so inherited services can create instances as they need to do it. 
Yeah, you told that you don't want to do it but I don't see other way to provide proper generic implementations.

sbespalov:
@fuss86 can you please provide a Test that requested [above](https://github.com/strongbox/strongbox/pull/987#issuecomment-459197223)?

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactCoordinatesService.java
sbespalov: :+1:

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.persistence.OneToMany;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+@MappedSuperclass
+public abstract class ArtifactGroup
```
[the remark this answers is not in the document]
fuss86: I don't need it, see [here](https://github.com/strongbox/strongbox-db/pull/5/files)

fuss86:
> @fuss86 can you please provide a Test that requested above?

Done, see [here](https://github.com/strongbox/strongbox/pull/987/commits/f32a576a30bf6ad2057170319fb8d58876945a15)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactEntryServiceImpl.java
fuss86: I like most expressed method names so I agree :)

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
fuss86: `name` dropped

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
fuss86: I still would like to avoid this unless it is necessary. We can do it if needed on later time.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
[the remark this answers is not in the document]
sbespalov: there is `name` property for `ArtifactGroup` class there.
anyway, I think we should have `ArtifactGroup.name` property because it makes real sense.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
sbespalov: Sorry, but it's seems that we should have it now.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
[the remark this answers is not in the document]
fuss86: > there is `name` property for `ArtifactGroup` class there.

I'll drop it since I don't need it. I forgot to do it.

>  anyway, I think we should have `ArtifactGroup.name` property because it makes real sense.

I don't see any sense right now since it is useless completely right now in the current implementaion.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
fuss86: I'm against of doing it, because current implementation may be used in generic way. If there are any problems with it in the future, we will change it. Right now it perfectly suits our needs any I believe it may fit to any others.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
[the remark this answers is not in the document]
fuss86: If you feel that it may be good to have it one day, we will add it one day. It is useless right now.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/impl/ArtifactGroupServiceImpl.java
carlspring: Guys, what seems to be the problem here?
@sbespalov : How would you like it to look?

fuss86: Done in https://github.com/strongbox/strongbox/pull/987/commits/2dc51fe56eefda5c5822ec9940c0f9f8007b8ec5

fuss86:
Is there anything to add here, @sbespalov ?

fuss86:
@carlspring I think you can review and merge this PR. But first we should merge https://github.com/strongbox/strongbox-db/pull/5

strongbox-data-service/pom.xml
```diff
@@ -79,7 +79,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>strongbox-db-liquibase</artifactId>
-            <version>${project.version}</version>
+            <version>PR-5-1.0-SNAPSHOT</version>
```
carlspring: Usually, the accepted (industry standard) Maven convention for this is `${baseVersion}-${issueKey}-SNAPSHOT`. For example: `1.0-SB-1234-SNAPSHOT`.

This is just for future reference and it doesn't make sense to change it right now.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/domain/ArtifactGroup.java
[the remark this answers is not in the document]
carlspring: Excuse me for joining in so late in this discussion. @fuss86: Would you mind explaining how it's used right now? I've had a look at the other pull request, but I'm not really sure I get it.

sbespalov:
superseded by https://github.com/strongbox/strongbox/pull/1026

PR #1004, opened 2019-02-03
Issue-1001: CronTaskController is unable to return list of tasks.

### LINKED ISSUE #1001: CronTaskController is unable to return list of tasks
The `CronTaskController#getConfigurations` is unable to return the list of cron tasks - it returns `406 Could not find acceptable representation`. This is likely because of broken serialization to JSON. We need to create a test case which checks if the endpoint actually returns a valid response with at least one result inside. 

NOTE: The returned crontasks JSON should look like this:

```
{
   "crontasks": [
      {uuid: ...., name: ....., etc },
      {uuid: ...., name: ....., etc },
      ......
   ]
}
```

ptirador:
Fixes #1001.

The problem was the `getConfigurations` method, because it was mapped to `/api/configuration/crontasks/` instead of `api/configuration/crontasks`.


carlspring:
This pull request looks good to me! Thanks, @ptirador!

@steve-todorov, let me know, if it works for you and I'll merge it! Thanks! :)

steve-todorov:
The test case in `CronTaskControllerTest#getCronConfigurations` looks like it has a typo:

```
given().contentType(MediaType.APPLICATION_JSON_VALUE)
       .accept(MediaType.APPLICATION_JSON_VALUE)
```

This is incorrect, because this would require a `body` to be sent with the specified `Content-Type`. However, in the current situation you don't have a body and this header is incorrectly used here. It should only have the `Accept` header ([RFC-7231](https://tools.ietf.org/html/rfc7231#section-5.3.2)) which is then used to properly negotiate the response format of the returned data.

The thing is that I've tried to remove the `contentType` part and the test still passes, but the curl request doesn't:

```
curl -u auser:apassword -i -H "Accept: application/json" http://localhost:48080/api/configuration/crontasks
```

I'm not entirely sure what's the deal here, but it looks like it's related to the content negotiation somehow. Either the serialization is broken or the negotiation breaks somewhere for some reason and the test cases are not catching it.

ptirador:
@steve-todorov, tests have been solved, so this is ready to merge if there's no issues!

carlspring:
@ptirador : Would you mind rebasing this PR (and using rebase in the future, instead of merge for cleaner PR-s)? It's hard to see what the actual changes are like this, unless one goes over the commits one by one.

In the core team we've had some discussions about this and we've decided to switch to git rebasing.

ptirador:
@carlspring Sorry about it. What can I do now to fix this?

carlspring:
No worries!

You can just rebase off the `master` and then force push the branch and it should be fine. This way, the noise from the merges should disappear.

Thanks for all your help!

steve-todorov:
> @steve-todorov, tests have been solved, so this is ready to merge if there's no issues! 

I confirm, this seems to be working fine for me now! Thanks for your work. We can merge it as soon as you rebase :)

ptirador:
I've not been able to 'rebase' cleanly, I've had to 'merge' first and the `rebase`, but the noise from merges is still there...

sbespalov:
@ptirador you can make a new branch from commit that you started your branch, then "cherry pick" your commits there. After this you can make a "squash" of your commits so there will be just one commit.
If you do all above then rebase will be not so hard rather than with many commits.

steve-todorov:
Closed in favor of #1021

PR #1006, opened 2019-02-03
Issue-1003: Saving group repositories does not preserve the order.

### LINKED ISSUE #1003: Saving group repositories does not preserve the order
When you go to the UI and try to add a Group Repository, the order of the grouped repositories **MUST** be the way it's ordered - otherwise, this could lead to possible invalid `maven-metadata.xml` due to the way Strongbox would generate the metadata.

ptirador:
Fixes #1003.

When updated, the group repositories set was deserialized as `HashSet` by default. It's been forced to deserialize as a `LinkedHashSet`.


carlspring:
Thanks for this quick fix! :)

PR #1007, opened 2019-02-04
updated tarLongFileMode to posix to avoid issues with mac users

Forbob:
tarLongFileMode configuration sets the TarArchiver behavior on file paths with more than 100 characters length. 

GNU tar has restrictions on max value of UID/GUID in tar files. Consider using the more well defined POSIX tar format, which should support larger values. Use tarLongFileMode=posix in the assembly:single goal.

This will avoid such issues in the future for Mac users on IT-provisioned machines in which your user id and group id are very high.

known issue with maven assembly plugin configuration: https://maven.apache.org/plugins/maven-assembly-plugin/faq.html#tarFileModes

### DIFF
--- a/strongbox-distribution/pom.xml
+++ b/strongbox-distribution/pom.xml
@@ -333,7 +333,7 @@
                         <descriptor>src/main/assembly/distribution.xml</descriptor>
                     </descriptors>
                     <appendAssemblyId>false</appendAssemblyId>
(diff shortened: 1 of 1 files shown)


carlspring:
Hi,

Congratulations on your first pull and thank you for your help!

This looks good to me!

PR #1010, opened 2019-02-05
SB-1359 proper snapshot tests

fuss86:


steve-todorov:
Thanks for fixing these tests! :)

PR #1013, opened 2019-02-07
SB-1364 maven-metadata expires after 10 seconds

fuss86:


steve-todorov:
Thanks for the quick fix! :)

PR #1017, opened 2019-02-10
SB-1362 Fix StorageController bugs 

fuss86:


steve-todorov:
I've tested this and it works fine now. Thank you for fixing it!

Issue #944, opened 2018-11-24
Fix failing test in strongbox-web-integration-tests for NuGet Layout Provider


ptirador:
After the [Spring Boot upgrade to 2.1.0 version](https://github.com/strongbox/strongbox/issues/919) (see pull request #930), the `strongbox-web-integration-tests` module for NuGet Layout Provider is failing. The following error is shown:

```
Execute command[s]: 
mono /usr/local/bin/nuget-3.4.4.exe push Org.Carlspring.Strongbox.Examples.Nuget.Mono/1.0.0/Org.Carlspring.Strongbox.Examples.Nuget.Mono.1.0.0.nupkg -ConfigFile /workspace/nuget/target/it/NuGet.config
Pushing Org.Carlspring.Strongbox.Examples.Nuget.Mono.1.0.0.nupkg to 'http://localhost:48080/storages/storage-nuget/nuget-releases'...
  PUT http://localhost:48080/storages/storage-nuget/nuget-releases/
18:23:52.259 24-11-2018 | INFO  | qtp1592099291-20     | o.c.s.users.userdetails.StrongboxUserDetailService | Authorise under SpringSecurityUser{username='admin', enabled=true, authorities (count) = 58, accessModel=null, url='null'}
18:23:52.394 24-11-2018 | INFO  | qtp1592099291-20     | o.c.s.c.layout.nuget.NugetArtifactController       | Nuget push request: storageId-[storage-nuget]; repositoryId-[nuget-releases]
0
18:23:52.428 24-11-2018 | ERROR | qtp1592099291-20     | o.c.s.c.layout.nuget.NugetArtifactController       | Failed to process Nuget push request: storage-nuget:nuget-releases
org.apache.commons.fileupload.MultipartStream$MalformedStreamException: Stream ended unexpectedly
	at org.apache.commons.fileupload.MultipartStream.readHeaders(MultipartStream.java:540)
	at org.carlspring.strongbox.controllers.layout.nuget.NugetArtifactController.writePackagePart(NugetArtifactController.java:626)
	at org.carlspring.strongbox.controllers.layout.nuget.NugetArtifactController.extractPackageMultipartStream(NugetArtifactController.java:588)
	at org.carlspring.strongbox.controllers.layout.nuget.NugetArtifactController.putPackage(NugetArtifactController.java:438)
[...]

PR #1023, opened 2019-02-13
SB-1266: PropertiesBooter improvements

carlspring:
More fixes.

This issue is affected by #944.

Tasks:
* [x] Sort out the issues with the initialization of beans (currently worked around by setting `spring.main.allow-bean-definition-overriding=true`).
* [x] Revert the `@SpringBootTest` changes in the `strongbox-web-core`.
* [ ] Re-work the `MutableStorage` to not use `System.getProperties`, but the Spring Boot Environment instead.
* [x] Sort out the `strongbox.vault_IS_UNDEFINED` issue. This is related to the logs and a logging property which is somehow not set, or not set on time.
* [ ] Investigate any duplicate initializations of Spring Boot showing multiple banners like these during the same testing phases (unit/integration):
```
  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v2.1.2.RELEASE)
```
  * Observed in:
    * [ ] `strongbox-authentication-providers`
    * [ ] `strongbox-storage-maven-layout-provider`
    * [ ] `strongbox-web-core`

Testing:
* [ ] Check, if there's an impact on the `strongbox-web-integration-tests`.
* [x] Check, if running `mvn spring:boot` in the `strongbox-web-core` still works.
* [x] Check, if running the `strongbox-distribution` still works.


strongbox-web-core/src/main/resources/application.properties
```diff
@@ -16,3 +20,5 @@ logging.dir=strongbox-vault/logs
 logging.config.file=strongbox/etc/logback-debug.xml
 java.io.tmpdir=strongbox-vault/tmp
 ehcache.disk.store.dir=strongbox/cache
+
+spring.main.allow-bean-definition-overriding=true
```
carlspring: Without this, it throws:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

The bean 'authenticationManager', defined in class path resource [org/carlspring/strongbox/configuration/StrongboxSecurityConfig.class], could not be registered. A bean with that name has already been defined in URL [jar:file:/home/carlspring/.m2/repository/org/carlspring/strongbox/strongbox-security-api/1.0-SNAPSHOT/strongbox-security-api-1.0-SNAPSHOT.jar!/org/carlspring/strongbox/security/managers/AuthenticationManager.class] and overriding is disabled.

Action:

Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true
```

@fuss86 , @ptirador : Could you maybe have a look?

ptirador: Have you tried erasing 

```
@Component
```
from `org.carlspring.strongbox.security.managers.AuthenticationManager` ?

carlspring:
No. Would you like to poke around?

strongbox-web-core/src/main/resources/application.properties
carlspring: @ptirador : No. Would you like to poke around?

ptirador: @carlspring: Solved. Everything seems working now...

carlspring: Thanks a lot! :)

carlspring: @ptirador : This seems to have broken the build...

ptirador: Hmm, it's weird, everything worked in my local environment...

carlspring:
@sbespalov , @fuss86 , @ptirador ,

We need to figure out the issues with the `spring.main.allow-bean-definition-overriding`. When this is not set to `true`, Spring fails to initialize.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/MutableStorage.java
```diff
@@ -87,7 +88,13 @@ public void setBasedir(String basedir)
 
     public void initDefaultBasedir(String id)
     {
-        Path basedirPath = Paths.get(System.getProperty("strongbox.storage.booter.basedir"));
+        //TODO: we should rework this to use SpringBoot environment instead of `System.getProperty` ASAP!!!
```
sbespalov: @carlspring the follow up task should be to rework this method.

carlspring: @ptirador ,

Would you like to look into this?

ptirador: @sbespalov Do you mean to inject `Environment` class from `Environment API`, and then use its `getProperty` method?

@Inject

sbespalov: @ptirador the thing is that we can't inject anything here, because this class is not under Spring context. So we should drop this `MutableStorage.initDefaultBasedir()` method and find appropriate solution how to set `MutableStorage.basedir` property from Spring environment.

sbespalov:
@carlspring 
> Sort out the strongbox.vault_IS_UNDEFINED issue. This is related to the logs and a logging property which is somehow not set, or not set on time.

this is also fixed with [this commit](https://github.com/strongbox/strongbox/pull/1023/commits/082dde9f2372077bc530797578ca8011458c2839)

carlspring:
Thanks, @sbespalov !

strongbox-resources/strongbox-common-resources/src/main/resources/logback-debug.xml
```diff
@@ -7,6 +7,8 @@
         <resetJUL>true</resetJUL>
     </contextListener>
 
+    <springProperty scope="context" name="loggingPath" source="strongbox.vault" defaultValue="."/>
```
fuss86: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-logback-extensions

> Because the standard logback.xml configuration file is loaded too early, you cannot use extensions in it. You need to either use logback-spring.xml

carlspring:
A massive thanks to everyone who helped out! :)

Issue #1030, opened 2019-02-15
java.lang.RuntimeException: Could not delete the role


carlspring:
# Task Description

The `org.carlspring.strongbox.controllers.configuration.security.authorization.AuthorizationConfigControllerTestIT.roleShouldNotBeDeleted` causes the following exception to be thrown:

```
18:41:32.327 15-02-2019 | ERROR | kJoinPool-1-worker-1 | o.c.s.c.c.s.a.AuthorizationConfigController        | Could not delete the role.
java.lang.RuntimeException: Could not delete the role.
	at org.carlspring.strongbox.controllers.configuration.security.authorization.AuthorizationConfigController.deleteRole(AuthorizationConfigController.java:141)
	at org.carlspring.strongbox.controllers.configuration.security.authorization.AuthorizationConfigController.deleteRole(AuthorizationConfigController.java:121)
	at org.carlspring.strongbox.controllers.configuration.security.authorization.AuthorizationConfigController$$FastClassBySpringCGLIB$$6a35eb7c.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:749)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:69)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:688)
	at org.carlspring.strongbox.controllers.configuration.security.authorization.AuthorizationConfigController$$EnhancerBySpringCGLIB$$ef12b464.deleteRole(<generated>)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #1041, opened 2019-02-16
Issue1030: java.lang.RuntimeException: Could not delete the role

adavid9:
#1030 This commit, improves handling of the exception and add logger, so that when you run the tests there is no longer NullPointerException stacktrace, and you get simple one line error: `could not delete the role.`


adavid9:
@carlspring I think this fix is ready for a review.

carlspring:
Thanks for looking into this! :)

PR #1056, opened 2019-02-21
Issue 965: Convert the strongbox configuration files from XML to YAML format

### LINKED ISSUE #965: Convert the strongbox configuration files from XML to YAML format
# Task Description

We would like to convert our configuration files from XML into YAML.

* [x] Re-work the `XmlFileManager` into `YamlFileManager`, so that we could start converting the XML files into YAML. The property names will have to be changed and they'll have to be camel case instead of using hyphens.
* `etc/conf/strongbox.xml`:
  * [x] Convert to `yaml`
  * [ ] Add a page called "The strongbox.yaml file" to the "User Guide" in the [strongbox-docs](https://github.com/strongbox/strongbox-docs).
* `etc/conf/strongbox-authorization.xml`:
  * [x] Convert to `yaml`
  * [ ] Add a page called "The strongbox-authorization.yaml file" to the "User Guide" in the [strongbox-docs](https://github.com/strongbox/strongbox-docs).
* `etc/conf/strongbox-cron-tasks.xml`:
  * [x] Convert to `yaml`
  * [ ] Add a page called "The strongbox-cron-tasks.yaml file" to the "User Guide" in the [strongbox-docs](https://github.com/strongbox/strongbox-docs).
* `etc/conf/strongbox-security-users.xml`:
  * [x] Convert to `yaml`
  * [ ] Add a page called "The strongbox-security-users.yaml file" to the "User Guide" in the [strongbox-docs](https://github.com/strongbox/strongbox-docs).
* [ ] Update the [Application Properties](https://strongbox.github.io/user-guide/application-properties.html) page.
* [ ] Update the [Routing Rules](https://strongbox.github.io/user-guide/artifact-routing-rules.html) page.

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

ptirador:
Fixes #965.

# Acceptance Test

* [x] Building the code and running `mvn spring-boot:run` in the `strongbox-web-core` still starts up the application correctly.
* [x] Running the [`strongbox-web-integration-tests`](https://github.com/strongbox/strongbox-web-integration-tests/) still run properly.
* [x] Running strongbox-against-strongbox works fine.

# Minimal TODO things:

* [x] Replace `strongbox-cron-tasks.xml` from `strongbox-cron/strongbox-cron-tasks/src/test/resources/etc/conf` with `yaml` version
* [x] Replace `strongbox-cron-tasks.xml` from `strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-nuget-layout-provider/src/test/resources/etc/conf` with `yaml` version
* [x] Replace `strongbox-cron-tasks.xml` from `strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/resources/etc/conf` with `yaml` version
* [x] Drop `org.carlspring.strongbox.xml.parsers.GenericParser` 
* [x] Replace package `org.carlspring.strongbox.xml` by `org.carlspring.strongbox.yaml` 
* [x] Synchronize `org.carlspring.strongbox.yaml.YamlFileManager` or `org.carlspring.strongbox.yaml.CustomYamlMapper`
* [x] Solve `"[nuget.org]"` problems with yaml
* [x] Use the same single instance of `strongbox.yaml` between modules for tests


strongbox-cron/strongbox-cron-api/src/main/resources/etc/conf/strongbox-cron-tasks.yaml
```diff
@@ -0,0 +1,37 @@
...
+    properties:
+      cronExpression: 0 0 1 * * ?
+      jobClass: org.carlspring.strongbox.cron.jobs.RebuildMavenMetadataCronJob
+  - uuid: c9e7eeef-4e97-4500-aa12-e714a246c2dc
+    name: Remove Timestamped Maven Snapshots Cron Job
+    oneTimeExecution: false
+    immediateExecution: false
+    properties:
+      cronExpression: 0 0 1 * * ?
+      jobClass: org.carlspring.strongbox.cron.jobs.RemoveTimestampedMavenSnapshotCronJob
```
carlspring: Cosmetic: Could you please add a new line at the end of such configuration resources? Thanks!

carlspring:
@ptirador,

Could I use this opportunity to tell you how 

**A W E S O M E**

you are...? :-D

ptirador:
Thanks so much :)
By the way, this is still a work in progress, it's not finished yet.
I have to think about the way of mapping correctly the DTOs with the corresponding annotations.

steve-todorov:
@ptirador this looks so pretty! :D 

img

strongbox-storage/strongbox-storage-core/pom.xml
```diff
@@ -64,7 +64,7 @@
                         <strongbox.vault>${project.build.directory}/strongbox-vault</strongbox.vault>
                         <strongbox.storage.booter.basedir>${project.build.directory}/strongbox-vault/storages
                         </strongbox.storage.booter.basedir>
```
carlspring: Cosmetic: Could you please make these properties one-liners? Thanks! :)

strongbox-web-core/src/test/resources/etc/conf/strongbox.yaml
```diff
@@ -0,0 +1,331 @@
+configuration:
```
carlspring: How is this resource different from the one in the `strongbox-storage-api-resources` module?

strongbox-configuration/src/main/java/org/carlspring/strongbox/yaml/YamlFileManager.java
```diff
@@ -1,53 +1,48 @@
...
-                    ServiceLoader<?> loader = ServiceLoader.load(clazz);
-                    loader.forEach(impl ->
-                                   {
-                                       contextClasses.add(impl.getClass());
-                                   });
-                }
-        );
-        parser = new GenericParser<>(contextClasses.toArray(new Class[0]));
+        myClazz = (Class<T>) ((ParameterizedType) getClass().getGenericSuperclass()).getActualTypeArguments()[0];
+        Stream.of(classes).forEach(ServiceLoader::load);
```
sbespalov: do we really need this for `YAMLMapper` ?

strongbox-configuration/src/main/java/org/carlspring/strongbox/yaml/YamlFileManager.java
```diff
@@ -79,12 +75,13 @@ public T read()
         {
             return null;
         }
-        
+
         try (InputStream inputStream = new BufferedInputStream(resource.getInputStream()))
         {
-            return parser.parse(inputStream);
+            JavaType type = yamlMapper.getTypeFactory().constructType(myClazz);
+            return yamlMapper.readValue(inputStream, type);
```
sbespalov: we probably can use `myClazz` directly here, without `JavaType` conversion.

ptirador:
Hi! When trying to deserialize the `strongbox-authorization.yaml` fille, it seems that there are problems deserializing `Set`  objects:
`Cannot deserialize instance of 'java.util.HashSet' out of START_OBJECT token`

I guess that the best option here would be to change the type from `Set` to `Map`?
For example, for this attribute:
`private Set<RoleDto> roles;`

The solution would be:
`private Map<String, RoleDto> roles;`
Where the role name would be the key.

The YAML file for this would be:
```
authorizationConfiguration:
  roles:
    ANONYMOUS_ROLE: => This would be the map key
      name: ANONYMOUS_ROLE
      description: Common anonymous user role
      privileges: ARTIFACTS_RESOLVE, SEARCH_ARTIFACTS
    USER_ROLE:  => This would be the map key
      name: USER_ROLE
      description: Common user role
      privileges: VIEW_USER

```

carlspring:
@ptirador ,

I don't think this looks very nice... 

What seems to be the problem exactly? @fuss86 , @sbespalov , @steve-todorov , do you have any advice, or suggestions on how we can tackle this?

carlspring:
@ptirador ,

Also, I was going to ask -- when you update the documentation, could you please add a few links to some YAML tutorials, or crash courses, (to help quickstart people who aren't used to the, or familiar enough with the YAML format)? Thanks! :)

strongbox-web-core/src/test/resources/etc/conf/strongbox-cron-tasks.xml
```diff
@@ -9,8 +9,6 @@
         <properties>
             <cronExpression>0 0 1 * * ?</cronExpression> <!-- mandatory -->
             <jobClass>org.carlspring.strongbox.cron.jobs.ClearRepositoryTrashCronJob</jobClass> <!-- mandatory -->
-            <!-- <storageId>storage-common-proxies</storageId> optional -->
-            <!-- <repositoryId>maven-central</repositoryId> optional -->
```
fuss86: Why do we still need **strongbox-cron-tasks.xml** for tests ? Shouldn't we drop it ?

strongbox-web-core/src/test/resources/etc/conf/strongbox.xml
```diff
@@ -296,7 +296,9 @@
                             data-store="file-system"
                             layout="npm"
                             type="proxy">
-                    <remote-repository url="https://registry.npmjs.org/"/>
```
fuss86: Shouldn't we drop this file too ?

fuss86:
> @ptirador ,

> I don't think this looks very nice...

> What seems to be the problem exactly? @fuss86 , @sbespalov , @steve-todorov , do you have any advice, or suggestions on how we can tackle this?

@ptirador, could you please push all your changes and let us know how to reproduce the problem you reported ?

strongbox-web-core/src/test/resources/etc/conf/strongbox-cron-tasks.xml
carlspring: Yes, I believe we should be able to do so.

strongbox-web-core/src/test/resources/etc/conf/strongbox.xml
carlspring: Yes, I believe we should be able to do so.

carlspring:
Yeah, if you could specify what your error is exactly, we could try and work it out, or ask on Stackoverflow.

sbespalov:
> When trying to deserialize the `strongbox-authorization.yaml` fille, it seems that there are problems deserializing `Set` objects:
> `Cannot deserialize instance of 'java.util.HashSet' out of START_OBJECT token`

@ptirador  have you tried to make it array like below?
```
roles: 
    - description: Common anonymous user role
      name: ANONYMOUS_ROLE
      privileges: 
        - name: ARTIFACTS_RESOLVE
        - name: SEARCH_ARTIFACTS
    - description: Common user role
      name: USER_ROLE
      privileges: 
        - name: VIEW_USER
    - description: Deployment role
      name: CUSTOM_ROLE
      privileges: 
        - name: Deploy
```

steve-todorov:
I was about to suggest what @sbespalov have said. His example could simplified even further:

```
roles: 
    - description: Common anonymous user role
      name: ANONYMOUS_ROLE
      privileges: 
        - ARTIFACTS_RESOLVE
        - SEARCH_ARTIFACTS
      privileges: [ ARTIFACTS_RESOLVE, SEARCH_ARTIFACTS ] (this is just an alternative way to write the same as above; it is considered to be an array of string privileges :) )
    - description: Common user role
      name: USER_ROLE
      privileges: 
        - VIEW_USER
    - description: Deployment role
      name: CUSTOM_ROLE
      privileges: 
        -  Deploy
```

carlspring:
@ptirador ,

I've rebased and force pushed! I hope this helps! :)

ptirador:
Hello,

Until now all YAML files have been done (except a few issues with the last one):

- [x] `strongbox-authorization.yaml`. Its parent class `org.carlspring.strongbox.authorization.dto.AuthorizationConfigDto, its members and children classes have been correctly annotated. 

- [x] `strongbox-cron-tasks.yaml`. Its parent class `org.carlspring.strongbox.cron.domain.CronTasksConfigurationDto`, its members and children classes have been correctly annotated. 

- [x] `strongbox-security-users.yaml`. Its parent class `org.carlspring.strongbox.users.dto.UsersDto`, its members and children classes have been correctly annotated. 

- [ ] `strongbox.yaml`. Its parent class `org.carlspring.strongbox.configuration.MutableConfiguration`, its members and children classes have been correctly annotated.

There's a few tasks left for the last one.

**Task 1:**

- [ ] Wait for the [routing rules improvements PR](https://github.com/strongbox/strongbox/pull/1078) to be done and adapt the YAML and Java annotations to the associated model.

**Task 2:**
 The class `org.carlspring.strongbox.storage.repository.MutableRepository` has this member:
`private MutableCustomRepositoryConfiguration repositoryConfiguration;`

This interface `MutableCustomRepositoryConfiguration` (which belongs to `strongbox-configuration`) in order to be deserialized, needs to be annotated with its subtypes:

```
@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "type")
@JsonSubTypes({
        @Type(value = MutableMavenRepositoryConfiguration.class, name = "mavenRepositoryConfiguration"),
        @Type(value = MutableNpmRepositoryConfiguration.class, name = "npmRepositoryConfiguration"),
        @Type(value = MutableNugetRepositoryConfiguration.class, name = "nugetRepositoryConfiguration"),
        @Type(value = MutableRawRepositoryConfiguration.class, name = "rawRepositoryConfiguration")
})
public abstract class MutableCustomRepositoryConfiguration
        implements RepositoryConfiguration
{

    @JsonIgnore
    public abstract CustomRepositoryConfiguration getImmutable();

}
```

The problem is that I can't add this subtypes because each one belongs to a module where there would be circular dependencies.

**Task 3:**
Same as previous task, except for this is happening at another member from `MutableRepository`:
`private List<MutableCustomConfiguration> customConfigurations = new ArrayList<>();`

**Task 4:**
Same as previous task, except for this is happening at a member from `MutableRemoteRepository`:
`private MutableRemoteRepositoryConfiguration customConfiguration;`

Could you help me with these tasks (especially with Task 2, which is related to the foloowing ones)?
@fuss86 @sbespalov @carlspring @steve-todorov 

Thank you so much.

strongbox-web-core/src/test/resources/etc/conf/strongbox-cron-tasks.xml
ptirador: Done.

strongbox-web-core/src/test/resources/etc/conf/strongbox.xml
ptirador: Done.

fuss86:
Hi @ptirador,

could you please try serializing/deserializing it using current solution (`@JsonTypeInfo`) ? I believe it works for JSON now, since we tested our REST API (see `org.carlspring.strongbox.controllers.configuration.StoragesConfigurationControllerTestIT#testAddGetRepository` for example). See also `org.carlspring.strongbox.forms.configuration.CustomRepositoryConfigurationForm` and `@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "layout")`.

You can try to serialize/deserialize it to XML too. It is possible that it needs to be improved like the `*Form` classes (see also `org.carlspring.strongbox.forms.configuration.MavenRepositoryConfigurationForm` and related `@JsonTypeName("Maven 2")`

carlspring:
@ptirador : 

Regarding:
```
@JsonSubTypes({ @Type(value = MutableMavenRepositoryConfiguration.class,
                      name = "mavenRepositoryConfiguration"),
                @Type(value = MutableNpmRepositoryConfiguration.class,
                      name = "npmRepositoryConfiguration"),
                @Type(value = MutableNugetRepositoryConfiguration.class,
                      name = "nugetRepositoryConfiguration"),
                @Type(value = MutableRawRepositoryConfiguration.class,
                      name = "rawRepositoryConfiguration") })
```
Is it possible to make things work with a similar SPI implementation like the one I believe we had for `CustomConfiguration`-s...? I mean -- in our current implementation all the configurations for the different layout providers are the same. If they have a need for custom configurations, they all extend the `CustomConfiguration` class. I would recommend that you look for the `src/main/resources/META-INF/services/org.carlspring.strongbox.xml.repository.RepositoryConfiguration` file across the codebase and have a think about it in a similar way, if possible.

ptirador:
@fuss86: I managed it t work, but I explain it:

**First option:**
First, I tried using in the abstract class, just:
`@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "layout")`

And in the implementations:
`@JsonTypeName(Maven2LayoutProvider.ALIAS)`

BUT this showed an error like this:
`Could not resolve type id 'Maven 2' into a subtype of MutableCustomRepositoryConfiguration.`

Investigating, this could be solved by registering subtype(s) when configuring the `ObjectMapper`:
`registerSubtypes(new NamedType(MavenRepositoryConfiguration.class, Maven2LayoutProvider.ALIAS));`

But this registering forces me to move the `YamlFileManager` (which injects the ObjectMapper) to another module to avoid circular dependencies.

So I declined the solution.

**Second option:**
I tried using in the abstract class, just:
`@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "type")`

And no annotations in the implementations.

This leaves a YAML file with a type indicating the full qualified class name for the subtype, BUT IT WORKS.
For example:

```
repositoryConfiguration:
    type: org.carlspring.strongbox.xml.configuration.repository.MutableMavenRepositoryConfiguration
    indexingEnabled: true
```

carlspring:
I think we'll need to figure out a way for it to not use the class fully-qualified class name in such cases. Perhaps, there's a workaround. Perhaps what I recommended would work better? Would you like to give it a try?

sbespalov:
> Investigating, this could be solved by registering subtype(s) when configuring the ObjectMapper:

@ptirador  this could be solved like it already was with `ServiceLoader.load`, so you can load all implementation classes and then register them in `ObjectMapper` you have in `YamlFileManager`.

fuss86:
> So I declined the solution.

I think you should continue working on this solution :) See also @sbespalov note above.

> Second option:
> This leaves a YAML file with a type indicating the full qualified class name for the subtype, BUT IT 
> WORKS.

We can't rely on the fully qualified class names since sometimes we do the refactoring and things may be moved to some other package, sorry :(

ptirador:
Ok, it seems good idea, but how I register the subtypes without creating circular dependencies? If I do somehting like this:

```
registerSubtypes(new NamedType(MavenRepositoryConfiguration.class, Maven2LayoutProvider.ALIAS));
        registerSubtypes(new NamedType(NugetRepositoryConfiguration.class, NugetLayoutProvider.ALIAS));
        registerSubtypes(new NamedType(RawRepositoryConfiguration.class, RawLayoutProvider.ALIAS));
        registerSubtypes(new NamedType(NpmRepositoryConfiguration.class, NpmLayoutProvider.ALIAS));
```

I'll need to add dependencies to the pom.xml from `strongbox-configuration`, to import these classes., which will create circular dependencies...

carlspring:
If you use the SPI like I suggested, you will be able to workaround this.

carlspring:
Could you please rebase? :)

fuss86:
Yes, @carlspring spoke about the same thing, I just realized :)

carlspring:
It's far from ideal, but it's the best solution by far for our scenario right now.

ptirador:
Hi again,

I managed to deserialize the subtypes by using `@JsonTypeInfo` by name, and registering the subtypes in the mapper.

I created a `CustomYamlMapper` class in `strongbox-web-core`:

```
@Component
public class CustomYamlMapper
        extends YAMLMapper
{

    @PostConstruct
    public void postConstruct()
    {
        enable(SerializationFeature.WRAP_ROOT_VALUE);
        enable(DeserializationFeature.UNWRAP_ROOT_VALUE);
        setAnnotationIntrospector(new JacksonAnnotationIntrospector());
        setSerializationInclusion(JsonInclude.Include.NON_NULL);

        registerSubtypes(new NamedType(MutableMavenRepositoryConfiguration.class, "mavenRepositoryConfiguration"));
        // Other subtypes registration [...]
    }
}
```

The `YamlFileManager` in `strongbox-configuration` injects this custom YAML mapper:
```
@Inject
@Qualifier("customYamlMapper")
private YAMLMapper yamlMapper;
```
The thing is, when I try to execute some tests (for example, `RepositoryTest` in  `strongbox-storage-api`), this error occurs:
```
java.lang.IllegalStateException: Failed to load ApplicationContext

	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:125)
	[...]
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'configurationFileManager': Unsatisfied dependency expressed through field 'yamlMapper'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.fasterxml.jackson.dataformat.yaml.YAMLMapper' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@javax.inject.Inject(), @org.springframework.beans.factory.annotation.Qualifier(value=customYamlMapper)}
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:596)
	[...]	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:593)
	... 71 more

```

It seems that I can't inject this when running those tests... This would happen because `strongbox-storage-api` cannot access to injected beans in  `strongbox-web-core?
How can I solve this? Maybe it was not a good idea to inject the mapper...?
Please help me.

carlspring:
@ptirador :

I've added an acceptance test check-list, for when this is ready for review.

fuss86:
I'll try to help tomorrow :)

fuss86:
@ptirador I've rebased and fixed your problem with `YAMLMapper`. There are still couple of things to do. I've added a `Minimal TODO things:` to the description of this task.

ptirador:
@fuss86 Thanks so much for your help!

fuss86:
Hi @ptirador !,

 we are in the middle way to drop the `org.carlspring.strongbox.xml.parsers.GenericParser` 👍 

We are also discussing "Use the same single instance of `strongbox.yaml` between modules for tests" point, so this thing needs to wait :)

We are left with 20 failing tests so I believe this is the next issue to solve :)

ptirador:
Hi @fuss86, I've been trying `s-w-i-t` and seems fine, but with Maven is throwing an exception:
```
strongbox-web-integration-tests-maven_1_f206ff30b47a | org.carlspring.strongbox.storage.ArtifactStorageException: org.carlspring.strongbox.storage.validation.artifact.version.VersionValidationException: Cannot deploy a release artifact to a repository with a SNAPSHOT policy!
strongbox-web-integration-tests-maven_1_f206ff30b47a |  at org.carlspring.strongbox.services.ArtifactManagementService.performRepositoryAcceptanceValidation(ArtifactManagementService.java:324)
strongbox-web-integration-tests-maven_1_f206ff30b47a |  at org.carlspring.strongbox.services.ArtifactManagementService.validateAndStore(ArtifactManagementService.java:92)
[...]
strongbox-web-integration-tests-maven_1_f206ff30b47a | Caused by: org.carlspring.strongbox.storage.validation.artifact.version.VersionValidationException: Cannot deploy a release artifact to a repository with a SNAPSHOT policy!
strongbox-web-integration-tests-maven_1_f206ff30b47a |  at org.carlspring.strongbox.storage.validation.version.MavenSnapshotVersionValidator.validate(MavenSnapshotVersionValidator.java:83)
strongbox-web-integration-tests-maven_1_f206ff30b47a |  at org.carlspring.strongbox.services.ArtifactManagementService.performRepositoryAcceptanceValidation(ArtifactManagementService.java:318)
```
We got to figure this out.

carlspring:
@ptirado: Is it actually failing, or is it printing an exception? One of the tests is meant to print this, because it's checking it.

fuss86:
s-w-i-t failed, see [jenkins build](https://jenkins.carlspring.org/job/strongbox/job/strongbox-web-integration-tests-from-sources/93/display/redirect)

carlspring:
I remember there were some default initializations for the `Repository` class that were setting these values for allowing deployments and overwrite. Are these still there? It might be failing because of these.

strongbox-configuration/src/main/java/org/carlspring/strongbox/yaml/ObjectMapperSubtypes.java
```diff
@@ -0,0 +1,46 @@
...
+import org.carlspring.strongbox.yaml.repository.remote.MutableRemoteRepositoryConfiguration;
+
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+public class ObjectMapperSubtypes
```
carlspring: Cosmetic: I think this should be renamed to `ObjectMapperSubTypes`.

strongbox-configuration/src/main/java/org/carlspring/strongbox/yaml/ObjectMapperSubtypes.java
```diff
@@ -0,0 +1,46 @@
...
+{
+
+    public static final ObjectMapperSubtypes INSTANCE = new ObjectMapperSubtypes();
+
+    private static final Set<Class<?>> TYPES = ImmutableSet.of(CustomRepositoryConfiguration.class,
+                                                               MutableCustomRepositoryConfiguration.class,
+                                                               MutableRemoteRepositoryConfiguration.class);
+
+    private volatile Set<Class<?>> subtypes;
```
carlspring: Cosmetic: I think this should be renamed to `subTypes`.

strongbox-configuration/src/main/java/org/carlspring/strongbox/yaml/ObjectMapperSubtypes.java
fuss86: Isn't `subtype` a single word ? :)

carlspring: Actually, you're right. Sorry.

carlspring:
Guys,

I have tested `strongbox-against-strongbox` locally and it works fine!

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 14:46 min
[INFO] Finished at: 2019-04-02T04:27:39+01:00
[INFO] ------------------------------------------------------------------------
```

I would like to thank you all for pairing up and helping in whatever way you could! This has been a massive undertaking!

@ptirador, thank you for being brave enough to take such a colossal task!
@fuss86 and @sbespalov : thanks for providing Pablo with all your help, guidance and support! :)

PR #1060, opened 2019-02-22
Issue-1059: Drop the org.apache.tomcat:catalina:6.0.53 dependency from the strongbox-event-api

### LINKED ISSUE #1059: Drop the org.apache.tomcat:catalina:6.0.53 dependency from the strongbox-event-api
# Task Description

Drop the `org.apache.tomcat:catalina:6.0.53` dependency from the `strongbox-event-api` and apply any necessary changes to the code. This dependency should no longer be required, as we implemented this with Spring instead.

Affected classes:
- [ ] `EventExecutorFactoryBean`

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

adavid9:
Fixes #1059.

@sbespalov  
@fuss86 

Is this solution okay, or not? I decided to remove everything what is connected with `catalina` because if the there are no libraries related to it, maybe there is no sense to hold it. I ran integration tests, and everything built without any issues. Service from `strongbox-distribution` is also working and I was able to manage the `strongbox` without any other issues. If that solution is incorrect, just let me know how should it looks like and then I will improve this code.

### DIFF
--- a/strongbox-event-api/pom.xml
+++ b/strongbox-event-api/pom.xml
@@ -118,11 +118,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.tomcat</groupId>
-            <artifactId>catalina</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-all</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


carlspring:
Great! Thanks for reviewing and thanks, @adavid9 ! :)

PR #1068, opened 2019-02-24
Issue-840: Create an RPM distribution

### LINKED ISSUE #840: Create an RPM distribution

carlspring:
Fixes #840. Done by @v3rtumnus!

Rebased @v3rtumnus 's #855.

@v3rtumnus ,

Sorry for getting back to this with such a big delay (it's not typical for our project), but we had more pressing problems with our Spring Boot migration, as well as some issues with the parallelization of our tests. I have rebased your original pull request and pushed it to a different branch.

@steve-todorov , @v3rtumnus : Would you like to try this out?

@steve-todorov : We will have to come up with a build process in Jenkins that either runs separately on a CentOS/Opensuse node after `strongbox` has been built, or make the Jenkins nodes have `rpmbuild` on them.

__Tasks:__
* [x] Rebase #855 onto the `master`.
* [x] Apply any required fixes to the rebased code, caused by changes from the `master`.
* [x] Update the version of the `maven-assembly-plugin` in the `strongbox-parent` to `3.1.1`.
* [x] Make the `rpm-maven-plugin` output the `rpm` under `target`, (instead of `target/rpm/strongbox-distribution/RPMS/noarch`).
* [x] Attach the produced `rpm` as a Maven artifact.
* [x] Fix the Maven plugin version warnings.
* [x] ~Drop the timestamp from the RPM's version.~ [Not possible].
* [x] Re-structure the `rpm` to use `/opt`.
* [x] The `lib` directory contains a ton of dependencies which shouldn't be there.
* [x] Add a the `rpm` build in a Maven profile, which is triggered by invoking `mvn clean install -Pbuild-rpm`.
* [x] ~Make the `finalName` of the RPM follow the RPM naming when deploying to Github.~ [Not possible]
* [x] Sort out variable comments raised by @steve-todorov.

__Useful `rpm` commands:__
* List installed packages:
```
rpm -qa 
```
* Check if the `strongbox` package is installed:
```
rpm -qa | grep strongbox
```
* Install package:
```
rpm -ivh strongbox-distribution-1.0-SNAPSHOT.rpm
```
* List package contents
```
rpm -qlp strongbox-distribution-1.0-SNAPSHOT20190309035450.x86_64
```
* Remove package:
```
rpm -e strongbox-distribution-1.0-SNAPSHOT20190309035450.x86_64
```


steve-todorov:
There are a few things which need to be changed here:

1. `/usr/local/strongbox` is probably not the best place to put the distribution. According to [Linux Foundation](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html)

   > No other directories, except those listed below, may be in /usr/local after first installing a FHS-compliant system.

   If we are going to be putting `Strongbox` under `/usr/local/` it will need to be under `/usr/local/bin/strongbox`, `/usr/local/etc, `/usr/local/lib`, etc.

   In my opinion, a better place to place the distribution under `/opt/strongbox` ([link](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html))

2. I suggest a few changes to the `.service`:

  ```
[Unit]
Description=Strongbox Repository Manager
Documentation=https://strongbox.github.io/
After=network-online.target firewalld.service
Wants=network-online.target

[Service]
Type=simple
User=strongbox
Environment=STRONGBOX_HOME=/opt/strongbox
Environment=STRONGBOX_VAULT=/opt/strongbox-vault
Environment=STRONGBOX_PORT=48080
ExecStart=/opt/strongbox/bin/strongbox console
ExecStop=/opt/strongbox/bin/strongbox stop
ExecReload=/bin/kill -s HUP $MAINPID
TimeoutSec=0
RestartSec=2
Restart=always

# Note that StartLimit* options were moved from "Service" to "Unit" in systemd 229.
# Both the old, and new location are accepted by systemd 229 and up, so using the old location
# to make them work for either version of systemd.
StartLimitBurst=3

# Note that StartLimitInterval was renamed to StartLimitIntervalSec in systemd 230.
# Both the old, and new name are accepted by systemd 230 and up, so using the old name to make
# this option work for either version of systemd.
StartLimitInterval=60s

[Install]
WantedBy=multi-user.target
  ```


---

Also, I have a feeling this might need to be separated into a new project:
1. How is this going to build on a Windows or MacOS ? (I doubt windows has `rpm-build` in a package manager somewhere? :D )
2. We will probably need to create some tests which check if the distribution actually works:
   - Does it start from `tar.gz` and does it work (i.e. `strongbox against strongbox build`) ?
   - Does it start from `zip` and does it work (i.e. `strongbox against strongbox build`) ?
   - Does it start from `rpm` and does it work (i.e. `strongbox against strongbox build`) ?
   - Does it start from `deb` and does it work (i.e. `strongbox against strongbox build`) ?

carlspring:
@v3rtumnus ,

Would you like to give this a try?

PR #1070, opened 2019-02-24
Issue 1046: Use the Spring Boot Actuator logger management instead of our own solution

### LINKED ISSUE #1046: Use the Spring Boot Actuator logger management instead of our own solution
# Task Description

After we integrate with Spring Boot Actuators (#1000), we should migrate our logging management to `spring-boot` (see https://docs.spring.io/spring-boot/docs/current/actuator-api/html/#loggers). We will drop `LoggingManagementController` and we need to update `LoggingManagementControllerTestIT` (also rename it) to test that are actuator config works fine in terms of loggers. We should also drop:

```
<dependency>
    <groupId>org.carlspring.logging</groupId>
    <artifactId>logback-configuration-rest</artifactId>
</dependency>
```

from project dependencies.

# Tasks
* Check which of the following methods have analogues in the Spring Boot Logging Actuator (and list them):
  * [x] `addLogger`
  * [x] `updateLogger`
  * [x] `deleteLogger`
  * [ ] `downloadLog`
  * [x] `liveLogStreaming`
  * [ ] `generateLogDirectoryListing`
* [ ] Re-work the controller so that it extends the Spring Boot Logging Actuator with the features that it doesn't have (that we need).
* [ ] Consider contributing the improved logging actuator to the Spring Boot project.

# Useful Links

* [Spring Boot: HOWTO Logging](https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html)
* [Spring Boot Actuators: Loggers](https://docs.spring.io/spring-boot/docs/current/actuator-api/html/#loggers)
* [Spring Boot: Configure Log Level](https://blog.codeleak.pl/2017/03/spring-boot-configure-log-level-in.html)

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

adavid9:
Fixes #1046 

I added simple `loggers` endpoint to the `.properties` file and modified the security for these endpoints, because the previous does not secure them for not authorized users. Tests that I wrote are simple tests, just to test the endpoints, in the future it should be improved, for now it is okay I believe, if there is anything what should be changed, let me know and I will do it.


strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/ActuatorEndpointsControllerTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+import static io.restassured.module.mockmvc.RestAssuredMockMvc.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.notNullValue;
+
+/**
+ * @author: adavid9
+ */
+
+
```
carlspring: Cosmetic: Please, remove these empty lines above. :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/LoggingManagementController.java
```diff
@@ -1,287 +0,0 @@
-package org.carlspring.strongbox.controllers;
```
carlspring: I don't think we should be dropping the controller completely, as there's functionality that the Spring Boot logging actuators don't have. For example, we have directory browsing, download entire logs, download config file, etc.

Could you please make an item list and check which of the methods have analogues in Spring Boot's implementation and which ones don't so that we could come up with an action plan?

Perhaps we could re-work our code and extend the Spring Boot Logging Actuator and contribute it back to the Springframework folks...?

fuss86:
Hi @adavid9 , I'll continue work on this task, if you don't  mind :)

FYI @steve-todorov

fuss86:
moved to ant continued in https://github.com/strongbox/strongbox/pull/1440

adavid9:
@fuss86 Hi Przemyslaw, sure go ahead.

steve-todorov:
@adavid9 Thanks for your work on this! I'm closing it in favor of #1440 :)

PR #1082, opened 2019-03-02
Issue-1071: Expose versioning information using Spring Boot Actuator

### LINKED ISSUE #1071: Expose versioning information using Spring Boot Actuator
# Task Description

We need to expose our versioning information using Spring Boot Actuators.

In Maven we have the following properties in place:
* `${project.version}` – a built-in Maven property, that shows what the version of the artifact that is being produced is.
* `${strongbox.revision}` – a property, exposd via the `org.codehaus.mojo:buildnumber-maven-plugin` that could be used to get the exact git revision.

These values could be replaced using Maven resource filtering of the `application.properties`.

# Tasks

* [x] Do properties replacement using Maven resource filtering against the `application.properties.
* [x] Expose the versioning information via Spring Boot Actuators.

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

adavid9:
Fixes #1071

### DIFF
--- a/strongbox-commons/src/main/java/org/carlspring/strongbox/booters/PropertiesBooter.java
+++ b/strongbox-commons/src/main/java/org/carlspring/strongbox/booters/PropertiesBooter.java
@@ -49,10 +49,10 @@ public class PropertiesBooter
     @Value("${strongbox.nuget.download.feed}")
     private boolean strongboxNugetDownloadFeed;
 
-    @Value("${strongbox.project.version}")
-    private String strongboxProjectVersion;
+    @Value("${strongbox.version}")
+    private String strongboxVersion;
 
-    @Value("strongbox.revision")
+    @Value("${strongbox.revision}")
     private String strongboxRevision;
 
     /**
@@ -187,14 +187,14 @@ public class PropertiesBooter
         this.strongboxNugetDownloadFeed = strongboxNugetDownloadFeed;
     }
 
-    public String getStrongboxProjectVersion()
+    public String getStrongboxVersion()
     {
-        return strongboxProjectVersion;
+        return strongboxVersion;
     }
 
-    public void setStrongboxProjectVersion(String strongboxProjectVersion)
+    public void setStrongboxVersion(String strongboxVersion)
     {
-        this.strongboxProjectVersion = strongboxProjectVersion;
+        this.strongboxVersion = strongboxVersion;
     }
 
     public String getStrongboxRevision()
--- a/strongbox-commons/src/main/java/org/carlspring/strongbox/booters/PropertiesBooter.java
+++ b/strongbox-commons/src/main/java/org/carlspring/strongbox/booters/PropertiesBooter.java
@@ -49,6 +49,12 @@ public class PropertiesBooter
     @Value("${strongbox.nuget.download.feed}")
     private boolean strongboxNugetDownloadFeed;
 
+    @Value("${strongbox.project.version}")
+    private String strongboxProjectVersion;
+
+    @Value("strongbox.revision")
+    private String strongboxRevision;
+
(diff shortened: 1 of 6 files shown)


adavid9:
@carlspring @steve-todorov This fail is not caused of that commit isn't it?

strongbox-commons/src/main/java/org/carlspring/strongbox/booters/PropertiesBooter.java
```diff
@@ -181,4 +187,19 @@ public void setStrongboxNugetDownloadFeed(boolean strongboxNugetDownloadFeed)
         this.strongboxNugetDownloadFeed = strongboxNugetDownloadFeed;
     }
 
+    public String getStrongboxProjectVersion() {
```
carlspring: Cosmetic: Could you please correct the indentation here and across the entire pull? Thanks! :)

carlspring:
Thanks for implementing this! It's exactly according to spec! :)
I'll merge it once the tests in the `master` have been sorted out.

carlspring:
Could you please rebase? This should fix the build.

adavid9:
@carlspring done ;)

carlspring:
One last thing... I was thinking that we should have a simple test that checks if the version and revision strings have been replaced (are not null), because things happen and during refactorings it is possible that we might break it. Would you mind adding check for this?

adavid9:
Of course I write simple test that checks that these values are not null ;)

carlspring:
@adavid9 : It looks like you'll have to rebase and resolve the merge conflict.

strongbox-commons/src/main/resources/application.properties
```diff
@@ -25,6 +27,9 @@ ehcache.disk.store.dir=strongbox/cache
 
 spring.servlet.multipart.enabled=false
 
+info.strongbox.version=${strongbox.version}
+info.strongbox.revision=${strongbox.revision}
```
carlspring: What's the purpose of these `info.*` variables? Don't we already have them a few lines above?

adavid9: It is necessary for the info actuator endpoint, isn't it? Without this properties, it does not work.

steve-todorov:
LGTM and it works (tested) :)

strongbox-commons/src/main/resources/application.properties
carlspring: Can't the actuator just use `strongbox.version` and `strongbox.revision`? I mean -- why do you need to have two sets of properties?

adavid9: Without that there is an exception and it does not work. I think that is necessary to leave it like that.

carlspring: I'm sorry, but could you please investigate this? We can't have two sets of properties. This will confuse people. We should only have the `strongbox.version` and `strongbox.revision` ones.

@sbespalov, @fuss86, @steve-todorov : Am I missing something here?

steve-todorov: From what I've seen in the documentation, this is necessary because the `info` controller displays everything from the `info.*` properties. However, it should also be possible to create a custom `InfoContributor` component which returns the information from the `${strongbox.version}`.

carlspring: Sounds good!

adavid9: I will try to implement it today and also test it.

carlspring: Looks good, thanks! :)

carlspring:
Thanks for your help, @adavid9! :)

PR #1132, opened 2019-03-18
Issue 1096: Move dependency exclusions from children to strongbox-parent

### LINKED ISSUE #1096: Cleanup the pom.xml files from dependency exclusions and extract them to the strongbox-parent
# Task Description

We have `pom.xml` files that have `<exclusions>`. These will have to be migrated to the [`strongbox-parent`](https://github.com/strongbox/strongbox-parent/), where this makes sense.

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

LeifBrooks:
Fixes #1096 

Move dependency exclusions from children POMs to strongbox-parent.

Remove redundant/unused dependencies. For example, spring-core and spring-beans are both covered by spring-context, so they can safely be removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>org.carlspring.strongbox</groupId>
         <artifactId>strongbox-parent</artifactId>
-        <version>1.0-SNAPSHOT</version>
+        <version>1.0-PR-14-SNAPSHOT</version>
         <relativePath/>
     </parent>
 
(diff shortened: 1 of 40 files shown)


LeifBrooks:
We will have to run the build again after strongbox-parent PR is merged

LeifBrooks:
Added spring-boot-starter-web back to springbox-web-core. This is needed, and its exclusion should stay with it as no other module is using it.

carlspring:
Note: The guava vulnerability is a false positive.

PR #1164, opened 2019-03-30
Issue 1138: Drop the default test users

evrimnurcelik:
- Removed test users developer01, maven, user
- Updated the privileges of the "deployer" user
- Updated the unit tests and related classes to work accordingly with the user changes


strongbox-rest-client/src/main/java/org/carlspring/strongbox/client/RestClient.java
```diff
@@ -33,37 +33,6 @@
 
     private static final Logger logger = LoggerFactory.getLogger(RestClient.class);
 
-
-    public static RestClient getTestInstance()
-    {
-        return getTestInstance("maven", "password");
-    }
-
-    public static RestClient getTestInstanceLoggedInAsAdmin()
```
carlspring: You will have to leave this one.

strongbox-rest-client/src/main/java/org/carlspring/strongbox/client/RestClient.java
```diff
@@ -33,37 +33,6 @@
...
-    {
-        return getTestInstance("admin", "password");
-    }
-
-    public static RestClient getTestInstance(String username,
```
carlspring: And this one.

strongbox-security/strongbox-user-management/src/test/java/org/carlspring/strongbox/users/service/UserServiceTest.java
```diff
@@ -240,61 +240,6 @@ public void testThatUserNameIsUnique()
                 Collectors.toList()).size(), CoreMatchers.equalTo(1));
     }
 
-    @Test
-    public void testPrivilegesProcessingForAccessModel()
```
carlspring: I think this test will have to be re-worked. Perhaps you can create a new user and assign them the priviledges required below and improve it a bit. I don't think we should drop it.

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenArtifactControllerTest.java
```diff
@@ -1095,29 +1095,6 @@ private boolean checkSnapshotVersionExistsInMetadata(Metadata versionLevelMetada
...
     }
 
-    /**
-     * User developer01 does not have general the ARTIFACTS_RESOLVE permission, but it's defined for single 'releases'
-     * repository. So because of dynamic privileges assignment they will be able to get access to artifacts in that
-     * repository.
-     */
-    @Test
-    @WithUserDetails("developer01")
-    public void testDynamicPrivilegeAssignmentForRepository()
```
carlspring: I think this test will need to be re-worked as well, instead of being dropped. You will have to create the user and carry out the operations as them. However, the `@WithUserDetails(...)` won't work that way, so you'll need to figure out how to do it differently.

carlspring:
A couple of things:
* Could you please rebase to our latest `master`? (See [this](https://strongbox.github.io/developer-guide/git/rebase-vs-merge.html) article)
* The username under which you're committing and pushing to Github seems to be mis-configured. You will have to look into how to configure your git settings properly, amend the author of the commits and then force push back to your branch. (When your user is properly set up, you see your avatar when against your commits in the commit history, otherwise it's a default grey avatar). If you don't fix this, your Github profile will not be associated with these commits.

PR #1159, opened 2019-03-28
Storage.basedir configuration improvements

sbespalov:
Fixes #1129 
Contains:
- `TestRepository` and `TestArtifact` annotations improvements
- `MutableStorage.initDefaultBasedir` method dropped
- `Storage.getBasedir()` usage dropped from tests
- partially fix for #1165

PR #1174, opened 2019-04-02
master tests hotfix

fuss86:


strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenSearchControllerTest.java
```diff
@@ -31,99 +32,84 @@
 public class MavenSearchControllerTest
```
sbespalov: same changes was done within [#1159](https://github.com/strongbox/strongbox/pull/1159) , so I suggest to wait it instead

carlspring:
Thanks, @fuss86! I'll close this in favour of #1159 , as discussed.

PR #1127, opened 2019-03-17
Issue-1124: Restore the generation of our Swagger documentation

carlspring:
Fixes #1124.

- Restored the Maven profile.
- Restored the Swagger templates.

PR #1175, opened 2019-04-03
Add springfox swagger generation support

### LINKED ISSUE #1124: Restore the generation of our Swagger documentation
# Task Description

At the moment, our Swagger documentation is no longer being generated. This used to be done in the [`strongbox-webapp`](https://github.com/strongbox/strongbox/), check [here](https://github.com/strongbox/strongbox-webapp/blob/master/pom.xml#L204). However, with our recent migration to Spring Boot, we dropped the need for the [`strongbox-webapp`](https://github.com/strongbox/strongbox/) and the [`strongbox-assembly`](https://github.com/strongbox/strongbox-assembly/), as our whole application is, in essence, in the `strongbox-web-core` (which is where `strongbox-web-core-${project.version}-spring-boot.jar` is being produced; this is then packaged into the `strongbox-distribution`).

This task is a sub-task of #1123 .

# Tasks

We need to:
* [x] Restore the Maven [Swagger generation profile](https://github.com/strongbox/strongbox-webapp/blob/master/pom.xml#L204) and add it to the `strongbox-web-core`'s `pom.xml`.
* [x] Restore the Swagger templates from [here](https://github.com/strongbox/strongbox-webapp/tree/master/src/main/swagger/templates) to `strongbox-web-core/src/main/swagger/templates`.
* [ ] Figure out where to package the produced Swagger documentation to, so that Spring Boot would be able to display it properly.

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

cbrianhill:
This PR fixes #1124 by doing the following:

- Adding springfox swagger and swagger-ui dependencies
- Configuring the metadata for the Swagger docket
- Leveraging the existing Swagger annotations for the actual API docs (already in place from previous implementation)
- Adding an integration test to ensure both swagger-ui and the swagger API docs are served by the application

This does not implement the customized styling that was previously there as part of #1127.


cbrianhill:
Merging https://github.com/strongbox/strongbox-parent/pull/18 will provide dependency info which should allow this to build.

cbrianhill:
This PR now allows the API docs to be served from our customary location at `/docs/rest/api.html`.  Some additional information is in https://github.com/strongbox/strongbox/issues/1124.

carlspring:
Thanks a million for your massive efforts with this! :)

Issue #1165, opened 2019-03-30
Migrate the Maven layout provider tests to use the new annotations


carlspring:
# Task Description

We recently implemented #1083 and #1084 and we now need to re-work tests listed below to use these annotations.

Please, note:
* This task is quite large, but the intention is for it to be done in chunks -- you can fix one, or more tests and create a pull request and link it to this task.
* Multiple people can work on this task. For this to work out properly, please add a comment, when you decide to pick some test(s) to work on, so that others know not to work on them.
* This task is good for new joiners who have sufficient experience with JUnit and testing and will quickly help them learn more about the project.

# What Needs To Be Done

* There should be no creation of repositories and test resources (such as artifacts) in initialization methods annotated with `@BeforeAll`, or `@BeforeEach`, such as, for example, `initialize()` methods.
* At the moment, the test methods are all creating their own test repositories and artifacts. The code for this will have to be dropped and replaced with the annotations.
* Drop the `removeRepositories()` methods.
* Drop the `cleanUp()` methods.
* ~Drop all fields that are names of storages and repositories. No two test methods/classes should share the same names of repositories, or artifacts.~ (@carlspring annotations have synchronization based on artifact and repository ID, so it should be safe not to use unique names this will only affect on concurrency factor)
* Each test method needs to generate its own test repository (or repositories) and test resources (typically, artifacts).

For an illustration of a properly migrated test, please have a look at the `RepositoryManagementTest`.

## To create repositories using annotations:
```
@ExtendWith(RepositoryManagementTestExecutionListener.class)
@Test
public void testWithRepository(@TestRepository(storageId = "storage0",
                                               repositoryId = "repo1",
                                               layout = MavenArtifactCoordinates.LAYOUT_NAME) 
                               Repository repository)
{
    System.out.println(repository.getId());
}
```
###### Note
* _`storage = "storage0"` is default value, so it can be omitted_

## To create indexed Maven repository using annotations:
```
@ExtendWith(RepositoryManagementTestExecutionListener.class)
@Test
public void testWithRepository(@TestRepository(repository = "indexed_repo1",
                                               layout = MavenArtifactCoordinates.LAYOUT_NAME,
                                               setup = MavenIndexedRepositorySetup.class) 
                               Repository repository)
{
    System.out.println(repository.getId());
}
```
## To create test artifacts using annotations:
```
@ExtendWith({ RepositoryManagementTestExecutionListener.class,
              ArtifactManagementTestExecutionListener.class })
@Test
[...]

PR #1229, opened 2019-04-18
Issue 1165: Migrate the MavenRepositoryManagementServiceImplTest and ArtifactMetadataServiceReleasesTest tests to use the new annotations

yugander-krishan-singh:
Added `@TestRepository` and `@TestArtifact` annotations in test classes `MavenRepositoryManagementServiceImplTest.java` and `ArtifactMetadataServiceReleasesTest.java`

Fixes for #1165.

### DIFF
--- a/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactMetadataServiceReleasesTest.java
+++ b/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactMetadataServiceReleasesTest.java
@@ -69,31 +69,7 @@ public class ArtifactMetadataServiceReleasesTest
     @Inject
     private RepositoryPathResolver repositoryPathResolver;
 
-
-    @BeforeAll
-    public static void cleanUp()
-            throws Exception
-    {
-        //cleanUp(getRepositoriesToClean());
-    }
-
-    @BeforeEach
-    public void initialize()
-            throws Exception
-    {
-        //createRepository(STORAGE0, REPOSITORY_RELEASES, RepositoryPolicyEnum.RELEASE.getPolicy(), false);
-    }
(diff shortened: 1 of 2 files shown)


carlspring:
@yugander-krishan-singh : Thanks for these fixes!

Could you please also do it for `MavenRepositoryManagementServiceImplTest.testCreateAndDelete`, or would it not work...?

yugander-krishan-singh:
> @yugander-krishan-singh : Thanks for these fixes!
> 
> Could you please also do it for `MavenRepositoryManagementServiceImplTest.testCreateAndDelete`, or would it not work...?

@carlspring I tried for `MavenRepositoryManagementServiceImplTest.testCreateAndDelete` but then I am getting the following error-
`org.springframework.context.ApplicationContextException: Failed to close context.; nested exception is java.io.IOException: Failed to close following repositories: [rmsi-releases-1]`

The reason is that @TestRepository annotation creates the repo in the beginning and then deletes it once the test method execution ends.

Now in `testCreateAndDelete` method the repository is deleted before the method ends -
```
closeIndexer(STORAGE0 + ":" + REPOSITORY_RELEASES_2 + ":" + IndexTypeEnum.LOCAL.getType());

        getRepositoryManagementService().removeRepository(STORAGE0, REPOSITORY_RELEASES_2);

        assertFalse(repositoryDir.exists(), "Failed to remove the repository!");
```
Now when the method execution ends due to @TestRepository annotation it agains tries to delete the same repo but since it has already been deleted it throws the error mentioned above.

I can use the annotation but then I need to create the repository once it is deleted and asserted so that @TestRepository finds the repo and deletes it. I have tried it and it works. But then I didn't see any point in doing so.
IMHO since the method is named createAndDelete as it tests creation and deletion of repository, we should not use annotation and let the existing behavior be.
Kindly let me know your thoughts.

yugander-krishan-singh:
> @yugander-krishan-singh : Thanks for these fixes!
> 
> Could you please also do it for `MavenRepositoryManagementServiceImplTest.testCreateAndDelete`, or would it not work...?

@carlspring I have made the changes. Pls check.

carlspring:
Thanks for looking into this, @yugander-krishan-singh! :)

PR #1236, opened 2019-04-20
Allow swagger API to be rendered when running from JAR file

### LINKED ISSUE #1230: Swagger documentation is not working under strongbox-distribution
# Task Description

After merging #1124 we noticed that it works only when you start it via `spring-boot:run` but doesn't work in the `strongbox-distribution`. This needs to be investigated and fixed.

Steps to reproduce:

```
$ mvn clean install
$ cd strongbox-web-core && mvn spring-boot:run
$ curl -I http://localhost:48080/docs/rest/api.html (Status code 200, works fine)
```
---
```
$ mvn clean install
$ cd strongbox-distribution/target && \
  unzip strongbox-distribution-1.0-SNAPSHOT.zip && \ 
  ./strongbox-distribution-1.0-SNAPSHOT/strongbox-1.0-SNAPSHOT/bin/strongbox console
$ curl -I http://localhost:48080/docs/rest/api.html (Status code 404, not working)
```

cbrianhill:
This fixes #1230.  The resource handler I added in the original implementation used a mapping that would only work when run from a source tree.  I've changed the mapping to refer to the classpath-based location that works either from a source tree or from the JAR file.

### DIFF
--- a/strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
+++ b/strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
@@ -201,7 +201,7 @@ public class WebConfig
         registry.setOrder(-1);
 
         registry.addResourceHandler("/docs/**")
-                .addResourceLocations("/docs/")
+                .addResourceLocations("classpath:/META-INF/resources/docs/")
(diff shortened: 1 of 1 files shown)


carlspring:
Thanks for the quick fix! :)

Issue #1151, opened 2019-03-25
Test our RPM distribution and update our docs on how to use it


carlspring:
# Task Description

We recently started publishing an RPM distribution and we need to be properly tested and have the steps on how to install it documented.

# Tasks

The following tasks will need to be carried out:
* [x] Test the RPM `strongbox-distribution`.
* [x] Document the steps on how to install the RPM.

# Task Relationships

This task is a follow-up of #840.

# Help

* [Our chat](https://chat.carlspring.org/)
* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

michiboo:
Hi I would like to try tackle this issue!

carlspring:
Hi @michiboo,

Thanks for offfering to help! We need multiple people to test it and provide feedback, as well as to help polish our documentation!

@cbrianhill has also offered to help with this.

@michiboo, feel feel to join our 
[chat channel](https://chat.carlspring.org/) to discuss things further.

cbrianhill:
After testing, noted one issue, which was fixed with https://github.com/strongbox/strongbox/pull/1239.  Have just added some docs to explain how to use the RPM via https://github.com/strongbox/strongbox-docs/pull/19.

carlspring:
@cbrianhill ,

Would you mind also updating our wiki, as necessary? Thanks! :)

carlspring:
Thank you very much! :)

PR #1239, opened 2019-04-20
Add a JRE dependency to our RPM

cbrianhill:
This adds a dependency on the virtual RPM package `jre` so that it won't install without a JRE present on the system.  Identified this as part of #1151 testing.


carlspring:
Thank, @cbrianhill! :)

PR #1240, opened 2019-04-20
Issue 1237 -- correct base URL and API path in swagger docs

### LINKED ISSUE #1237: Swagger documentation is using the wrong base url
# Task Description

Swagger seems to be using a hard-coded base url which points to `http://carlspring.org/strongbox/api/**`. This URL needs to actually be the current domain under which you have opened the documentation (i.e. `http://localhost:48080/api/**`,  `http://domain.com/api` or whatever the domain is)

cbrianhill:
This fixes #1237 by removing the configured hostname for the API base URL (the system will infer it based on the request), and setting the API base path to `/api`.


carlspring:
Thanks for the quick fix! :)

PR #1269, opened 2019-05-08
Custom MVC mapping annotation for Artifact Controllers

### LINKED ISSUE #1268: Create custom MVC mapping annotation for Artifact Controllers
### Problem

Common Spring MVC `@RequestMapping` not supports multiple values mapping for same header name. So below example will not work as expected:
```
@RequestMapping(path = PypiArtifactController.ROOT_CONTEXT,
                headers = { "user-agent=Python-urllib/*", "user-agent=pip/*" })
```

### Solution

We need to implement custom mapping annotation which will be used to define custom MVC mapping for Artifact Controller classes:
```
@ArtifactRequestMapping(layout=PypiArtifactCoordinates.LAYOUT_NAME, userAgents={"Python-urllib/*", "pip/*" })
```

### How to do

We can make it similar with [this](https://stackoverflow.com/a/10336769/6825250)

sbespalov:
Fixes #1268 

- `@LayoutRequestMapping` annotation implemented
- `HeaderMappingFilter` and `HeaderMappingRegistry` dropped
- `contextBaseUrl` configuration fixed in tests


sbespalov:
@fuss86 this probably can be useful for your current task as well

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
```diff
@@ -131,10 +139,10 @@ protected String createMessage(HttpServletRequest request,
         return result;
     }
 
-    @Bean
-    HeaderMappingFilter headerMappingFilter()
+    @Override
+    protected RequestMappingHandlerMapping createRequestMappingHandlerMapping()
```
carlspring: The name sounds a bit weird... Wouldn't `RequestMappingHandler` be a better name for this?

sbespalov:
It called so in Spring.

ср, 15 мая 2019 г., 20:16 Martin Todorov <notifications@github.com>:

carlspring:
Thanks for these fixes! :)

Issue #528, opened 2018-01-22
HandlerMethodArgumentResolver to resolve Repository instance based on `{storageId}/{repositoryId}` path variables


fuss86:
We'd like to have a Controller request method argument resolver for `org.carlspring.strongbox.storage.repository.Repository` instances.

For example, consider existing `org.carlspring.strongbox.controllers.TrashController#undelete` method which looks like this:

```
@RequestMapping(value = "{storageId}/{repositoryId}/{path:.+}", method = RequestMethod.POST)
public ResponseEntity undelete(@PathVariable String storageId, @PathVariable String repositoryId, @PathVariable String path)
```
In many Controllers we have repeated pre-conditions validating existence of the repository for provided path variables. 

It'd be awesome if we could refactor above implementation to something like:

```
@RequestMapping(value = "{storageId}/{repositoryId}/{path:.+}", method = RequestMethod.POST)
public ResponseEntity undelete(@RepositoryMapping Repository repository, @PathVariable String path)
```

Implemented `HandlerMethodArgumentResolver` should validate:
* Repository existence (followed by storage existence check first, of course)
* If repository `Repository.isInService().`

If validation fails, we should throw some newly implemented Exception(s) and handle it/them using `@ControllerAdvice`/`@ExceptionHandler` spring features.

sbespalov:
@fuss86 I think this validation could also check that target repository is in service `Repository.isInService()`.

carlspring:
> @fuss86 I think this validation could also check that target repository is in service `Repository.isInService()`.

I think it should also, check if there is such a storage in the first place.

sbespalov:
@fuss86 also I have thoughts that we can provide `Repository` instance using something like `ThreadLocal` context which can be populated within `HeaderMappingFilter`, the thing is that we already have similar validation logic there. In this case `HandlerMethodArgumentResolver` will just check these `Repository` instance in `ThreadLocal` context.

fuss86:
> @fuss86 I think this validation could also check that target repository is in service Repository.isInService().

Correct, updated ticket description. Thanks.

> I think it should also, check if there is such a storage in the first place.

Of course.

> we can provide Repository instance using something like ThreadLocal context 

I like this idea, in general ...

> which can be populated within HeaderMappingFilter

[...]

PR #1272, opened 2019-05-09
SB-1325 Proxy artifact resolution issue - minimizing risk

fuss86:


carlspring:
Does this make #528 obsolete?

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.web.method.HandlerMethod;
+import org.springframework.web.servlet.HandlerMapping;
+import static org.carlspring.strongbox.storage.metadata.MetadataHelper.MAVEN_METADATA_XML;
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+public class MavenArtifactControllerInterceptor
+        extends BaseArtifactControllerInterceptor
```
sbespalov: @fuss86 am I understand correctly that inheritance needed here just because of `BaseArtifactControllerInterceptor.getStorage()` method? 
If so then I suggest to use `StoragesConfigurationManager.getStorage` in both interceptors instead.

Also we probably better to rename `BaseArtifactControllerInterceptor` into something more certain like `RepositoryInterceptor` or maybe `RepositoryRequestInterceptor`. 

Also we probably shouldn't use `Controller` keyword in `MavenArtifactControllerInterceptor` as well, because it's possible to have it not just for `MavenArtifactController` class.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.servlet.HandlerMapping;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+public class BaseArtifactControllerInterceptor
```
sbespalov: @fuss86 I think we should use it within `CustomRequestMappingHandlerMapping` (see `AbstractHandlerMapping.adaptedInterceptors`). It should be similar of how `HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE` works, so this interceptor should check `storageId`+`repositlryId` and if all checks passed then expose this variables as request attributes. These attributes will be used within `LayoutRequestCondition` to match the layout.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+            return false;
+        }
+        if (path.endsWith("/"))
+        {
+            response.sendError(BAD_REQUEST.value(), "The specified path should not ends with `/` character!");
+            return false;
+        }
+
+        final boolean mavenArtifactRequest = isMavenArtifactRequest(request.getMethod());
+        if (mavenArtifactRequest)
```
sbespalov: it will be less code complexity like below:
```
if (!mavenArtifactRequest)
{
    return true;
}
```

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+                                   "The specified path is invalid. Maven artifact artifactId not recognized.");
+                return false;
+            }
+            if (StringUtils.isBlank(mavenArtifact.getGroupId()))
+            {
+                response.sendError(BAD_REQUEST.value(),
+                                   "The specified path is invalid. Maven artifact groupId not recognized.");
+                return false;
+            }
+            if (!mavenArtifact.getArtifactId().startsWith(MAVEN_METADATA_XML))
```
sbespalov: we can reduce complexity here as well:
```
if (mavenArtifact.getArtifactId().startsWith(MAVEN_METADATA_XML)))
{
    return true;
}
```

strongbox-web-core/src/main/java/org/carlspring/strongbox/web/Constants.java
```diff
@@ -0,0 +1,14 @@
+package org.carlspring.strongbox.web;
+
+/**
+ * @author Przemyslaw Fusik
+ */
+public final class Constants
```
sbespalov: why this is not an interface?

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenArtifactControllerTest.java
```diff
@@ -682,27 +689,12 @@ public void whenRepositoryIsOutOfServiceWeShouldDisallowArtifactDeployment()
...
-    {
-        String path = "/storages/storage-common-proxies/maven-central/john/doe";
-        ExtractableResponse response = client.getResourceWithResponse(path, "");
-        assertEquals(response.statusCode(), HttpStatus.NOT_FOUND.value(), "Wrong response");
-    }
-
-    @Test
-    public void testNonExistingArtifactInNonExistingDirectory()
-    {
-        String path = "/storages/storage-common-proxies/maven-central/john/doe/who.jar";
```
sbespalov: we probably should make new test case for not found artifact with valid path.

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenArtifactControllerTest.java
```diff
@@ -682,27 +689,12 @@ public void whenRepositoryIsOutOfServiceWeShouldDisallowArtifactDeployment()
     }
 
     @Test
-    public void testNonExistingDirectoryDownload()
+    public void testDirectoryDownload()
```
sbespalov: we probably also need a test case with invalid file path, like it was `/storages/storage-common-proxies/maven-central/john/doe/who.jar` .

sbespalov:
@fuss86 this approach for resolving `SB-1325` looks really nice!

just need to integrate it with that we already have with `LayoutRequestMapping`.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
fuss86: ok, I've renamed `BaseArtifactControllerInterceptor` to `ArtifactRequestInterceptor` and `MavenArtifactControllerInterceptor` to `MavenArtifactRequestInterceptor`

> If so then I suggest to use StoragesConfigurationManager.getStorage in both interceptors instead.

done

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
fuss86: done

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
fuss86: done

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
fuss86: I like your idea, but:

1. `LayoutRequestCondition#getMatchingCondition` is used to determine the handler/method. Interceptors operate on the handler/method. Your suggestion changes the order of this flow.

2. Even if we try to do it (by not using interceptors) in the `CustomRequestMappingHandlerMapping#getMatchingMapping`, for example (so before  `LayoutRequestCondition#getMatchingCondition` ), then your ` HttpServletRequest` is wrapped by `HandlerMappingIntrospector.RequestAttributeChangeIgnoringWrapper` which has 
```
               @Override
		public void setAttribute(String name, Object value) {
			// Ignore attribute change...
		}
```

so maybe there is a reason Spring is doing it this way ?

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactControllerInterceptor.java
fuss86: > Also we probably better to rename BaseArtifactControllerInterceptor into something more certain like RepositoryInterceptor or maybe RepositoryRequestInterceptor.

`RepositoryRequestInterceptor` sounds the best, thanks

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenArtifactControllerTest.java
fuss86: there is already `MavenArtifactControllerTest#testNonExistingArtifactInExistingDirectory` and it is testing is just fine

strongbox-web-core/src/test/java/org/carlspring/strongbox/controllers/layout/maven/MavenArtifactControllerTest.java
fuss86: I restored `MavenArtifactControllerTest#testNonExistingArtifactInNonExistingDirectory` with `/storages/storage-common-proxies/maven-central/john/doe/who.jar` path but the error is 400 now, because of new interceptor validation.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
fuss86: However, I partially used your suggestion and `RepositoryRequestInterceptor` will be exposing `Repository` instance in the `HttpServletRequest`. It is however not used in `LayoutRequestCondition` as described above.

fuss86: Also, adding this interceptor as `AbstractHandlerMapping` interceptor is a bad choice, since for example `StoragesConfigurationController#addOrUpdateRepository` should not validate existence of the storage and repository.

I'm reverting it back.

sbespalov: @fuss86 thanks for having a look at this. I will also check your findings in this branch and get back with result.

sbespalov: @fuss86 seems you right about `RequestAttributeChangeIgnoringWrapper` but anyway the `HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE` exposed somehow despite the fact that at some point it goes through  `HandlerMappingIntrospector`. 
Anyway I suggest to work out it same way as Spring MVC does with other request attributes:
1. We should improve the `LayoutRequestCondition` so `getMatchingCondition` method will return new custom `RequestCondition` instance with `storage` and `repository` fields;
2. Override the `handleMatch` method in `CustomRequestMappingHandlerMapping` and expose the `storage` and `repository` variables from `RequestMappingInfo.customConditionHolder` into request attributes (`HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE` exposed within this method as well)
3. Use this attributes from request within `RepositoryRequestInterceptor` to check if requested storage and repository was found and if found check the `inSrvice`.

Also, this way you even don't need the `configurationManagementService` in `RepositoryRequestInterceptor` because it will check that `storage` and `repository` instances should be provided as request attributes, same way as it use provided `pathVariables`.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/MavenArtifactRequestInterceptor.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+public class MavenArtifactRequestInterceptor
+        extends HandlerInterceptorAdapter
+{
+
+    @Inject
+    private RepositoryPathResolver repositoryPathResolver;
+
+    @Inject
+    private ConfigurationManagementService configurationManagementService;
```
sbespalov: @fuss86 I mean that we should use `StoragesConfigurationManager` here instead of `ConfigurationManagementService`. There is shortcut methods `StoragesConfigurationManager.getStorage` and `StoragesConfigurationManager.getRepository` instead of `ConfigurationManagementService.getConfiguration().getStorage(storageId)`.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/RepositoryRequestInterceptor.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private final String repositoryIdPathVariableName;
+
+    private final String exposedRepositoryRequestAttributeName;
+
+    private final ConfigurationManagementService configurationManagementService;
+
+    public RepositoryRequestInterceptor(String storageIdPathVariableName,
+                                        String repositoryIdPathVariableName,
+                                        String exposedRepositoryRequestAttributeName,
```
sbespalov: @fuss86 what is the reason not to use constants directly instead of this parameters?

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
fuss86: > HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE exposed somehow despite the fact that at some point it goes through HandlerMappingIntrospector.

It is exposed later, I described the situation above

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/RepositoryRequestInterceptor.java
fuss86: You can use this interceptor for other PathVariables like `srcStorageId`/`srcRepositoryId` for example

sbespalov: ok, but this wouldn't be required in case of [last](https://github.com/strongbox/strongbox/pull/1272#discussion_r287205740) proposed solution

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
sbespalov: it was not clear, sorry :)
please share your thoughts about solution above

sbespalov:
@fuss86 I've just noticed one more thing, seems we can try to obtain the layout specific interceptors from application context with `AbstractHandlerMapping.detectMappedInterceptors()` method. So there will be no need to do it manually like you do with `MavenArtifactInterceptor`.

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/RepositoryRequestInterceptor.java
fuss86: but it still provides more flexibility for future cases like the one I noted above

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/BaseArtifactControllerInterceptor.java
fuss86: nevermind, done :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/interceptors/RepositoryRequestInterceptor.java
fuss86: nevermind, done :)

fuss86:
> @fuss86 I've just noticed one more thing, seems we can try to obtain the layout specific interceptors from application context with AbstractHandlerMapping.detectMappedInterceptors() method. So there will be no need to do it manually like you do with MavenArtifactInterceptor.

Please check latest commits if it matches your proposal, because I'm not sure

fuss86:
> Does this make #528 obsolete?

No, @carlspring, but will be useful for that task

sbespalov:
> Please check latest commits if it matches your proposal, because I'm not sure

@fuss86 many thanks for this

I've also pushed one small improvement to completely match the proposal

sbespalov:
@carlspring this is #LGTM :)

there will be no more comments from me here

carlspring:
Thanks for the hard work and for the teamwork! :)

PR #1391, opened 2019-07-24
Fix MacOS failure in StrongboxIndexerTest

cbrianhill:
# Pull Request Description

This fixes two failures in `StrongboxIndexerTest` on MacOS with APFS filesystem (our Travis-CI build hosts run HFS+).  The issue causing the failures is complex:

- The class `MinimalArtifactInfoIndexCreator` (https://github.com/apache/maven-indexer/blob/master/indexer-core/src/main/java/org/apache/maven/index/creator/MinimalArtifactInfoIndexCreator.java) in Apache's maven-indexer will associate the checksums from a Java artifact with the POM entry in the index if the Java artifact/checksum exist when the POM is added to the index.  I'm uncertain at this point whether this is intended behavior, but it does not seem to have changed recently.
- The comparator `TestArtifactContext.repositoryPathChecksumComparator` appears to have been implemented incorrectly.  Depending on the order in which paths are added to a collection, sorting using this comparator can yield different results.
- The order in which repository paths are added to the TreeMap in `TestArtifactContext.deployArtifact()` is determined by `Files.newDirectoryStream()` from the JRE, which does not guarantee any particular order.
- On Macs running APFS, the files are returned in an order such that the existing comparator will sort the JAR file ahead of the POM file, triggering the unexpected behavior.
- `StrongboxIndexerTest.indexerShouldBeCapableToSearchByFullSha1Hash` and `StrongboxIndexerTest.indexerShouldBeCapableToSearchByPartialSha1Hash` both fail because they return 2 total hits instead of the expected 1.  One of the hits comes from the POM file, the other comes 
from the JAR file.  This is because both index entries end up with the SHA1 hash that was originally generated from the JAR file.

# Acceptance Test

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes, please see #...
  * [x] No

  * [x] No

### DIFF
--- a/strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/artifact/TestArtifactContext.java
+++ b/strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/artifact/TestArtifactContext.java
@@ -215,7 +215,14 @@ public class TestArtifactContext implements AutoCloseable
 
         Files.delete(e.getValue());
     }
-    
+
+    /**
+     * This comparator will sort the paths in reverse alphabetical order, with
+     * checksum files ordered after other types of files.  This is necessary because
+     * Apache Maven Indexer's MinimalArtifactInfoIndexCreator will apply the checksum
+     * from a JAR file to the POM's entry in the index if the JAR file is added to the
+     * repository before the POM.
+     */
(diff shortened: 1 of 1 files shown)


sbespalov:
@cbrianhill  seems like a good catch, thanks! :+1:

carlspring:
Thank you for this great explanation of the problem and for actually fixing it, Brian! :)

Issue #1412, opened 2019-08-01
Expose logging configuration properties for s-w-i-t


sbespalov:
# Task Description
Currently there is no way to enable debug logging for s-w-i-t. 

# Tasks

The following tasks will need to be carried out:
* [x] Expose `logging.debug` property as `STRONGBOX_DEBUG` environment variable
* [x] Expose `logging.console.enabled` property as `STRONGBOX_LOG_CONSOLE_ENABLED` environment variable
* [x] Expose `logging.file.enabled` property as `STRONGBOX_LOG_FILE_ENABLED` environment variable

# Task Relationships

This task:
* Relates to: SB-1400

steve-todorov:
@sbespalov I was thinking that we should probably add a warning comments over these environment variables both in `strongbox/strongbox` and `strongbox/strongbox-web-integraion-tests`. The comment should mention that changing these will require changes in the opposite project and the pipeline code. If we ever decide to change this - it will for sure break everything down the line so it would be good to have it in mind. Maybe something like this would  be enough:

```
/** 
 * Changes to this env variable must also be applied to [strongbox/strongbox || strongbox/strongbox-web-integration-tests] and possibly the pipeline. 
 * Please ping us in the main channel to confirm.
 */
```

PR #1414, opened 2019-08-02
issues/1412: comments for logging configuration

sbespalov:
Related to #1412


strongbox-commons/src/main/resources/application.yaml
```diff
@@ -33,6 +33,7 @@ strongbox:
   server:
     database:
       path: ${strongbox.vault}/db
+# Changes to this proeprties must also be applied to [strongbox/strongbox-web-integration-tests] and possibly the pipeline. Please ping us in the main channel to confirm.
```
steve-todorov: @sbespalov shouldn't this be under the `setenv.conf` and `strongbox-distribution/pom.xml`? I was thinking we should add this comment at the places where the environment variables are actually being set/exposed.

sbespalov: @steve-todorov changes on `setenv.conf` and `strongbox-distribution/pom.xml` will not affect on anything :)
only this properties changes will affect on env variables

steve-todorov: Thanks for clarifying! :)

PR #1419, opened 2019-08-05
Make Hazelcast configurable

### LINKED ISSUE #1392: Make the Hazelcast settings configurable
# Task Description

Currently, the Hazelcast configuration is hard-coded in the `org.carlspring.strongbox.config.hazelcast.HazelcastConfiguration`. We need to be able to configure this via the `application.yaml` instead.

# Tasks

* [ ] Make the Hazelcast settings configurable via the `application.yaml`.
* [ ] ~~Make the Hazelcast settings configurable via the controller.~~ (this is currently not possible)

# Help

* Points of contact:
  * @carlspring
  * @fuss86
  * @sbespalov
  * @steve-todorov

cbrianhill:
# Pull Request Description

This pull request closes #1392

# Acceptance Test

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes, please see #...
  * [x] No

  * [ ] No

### DIFF
--- a/strongbox-commons/src/main/resources/application.yaml
+++ b/strongbox-commons/src/main/resources/application.yaml
@@ -66,7 +66,7 @@ cacheManagerConfiguration:
   groupConfig:
     name: strongbox
     password: password
-  enableMulticastConfig: true
+  enableMulticastConfig: false
   multicast:
     multicastGroup: 224.2.2.3
     multicastPort: 54327
--- a/strongbox-commons/src/main/resources/application.yaml
+++ b/strongbox-commons/src/main/resources/application.yaml
@@ -62,4 +62,31 @@ management:
       exposure:
         include: health,info,beans,metrics,trace,scheduledtasks,threaddump
       base-path: /api/monitoring
-
+cacheManagerConfiguration:
+  groupConfig:
+    name: strongbox
+    password: password
+  enableMulticastConfig: true
+  multicast:
+    multicastGroup: 224.2.2.3
+    multicastPort: 54327
+    multicastTimeoutSeconds: 2
+    multicastTimeToLive: 32
+    trustedInterfaces: en0,en1
+    loopbackModeEnabled: false
+  instanceId: strongbox
+  caches:
+    remoteRepositoryAliveness:
+      maxSizeLimit: 1000
+      maxSizePolicy: FREE_HEAP_SIZE
+      evictionPolicy: LFU
+    tags:
+      maxSizeLimit: 1000
+      maxSizePolicy: FREE_HEAP_SIZE
+      evictionPolicy: LFU
+    authentications:
+      cacheLocalEntries: true
+      evictionConfigSize: 1000
+      evictionConfigMaxSizePolicy: ENTRY_COUNT
+      invalidateOnChange: true
+      timeToLiveSeconds: 10
--- a/strongbox-data-service/src/main/java/org/carlspring/strongbox/config/hazelcast/HazelcastConfiguration.java
+++ b/strongbox-data-service/src/main/java/org/carlspring/strongbox/config/hazelcast/HazelcastConfiguration.java
@@ -2,16 +2,15 @@ package org.carlspring.strongbox.config.hazelcast;
 
 import org.carlspring.strongbox.data.CacheName;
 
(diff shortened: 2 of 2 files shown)


strongbox-commons/src/main/resources/application.yaml
```diff
@@ -62,4 +62,31 @@ management:
       exposure:
         include: health,info,beans,metrics,trace,scheduledtasks,threaddump
       base-path: /api/monitoring
-
+cacheManagerConfiguration:
+  groupConfig:
+    name: strongbox
+    password: password
```
carlspring: Could we have this as a non-plain-text value? @strongbox/core-developers : Any suggestions?

And... perhaps the `application.yaml` should only be viewable by the owner (we can control this in the `strongbox-distribution`'s assembly descriptor).

carlspring:
@cbrianhill : Thanks for this pull!

@strongbox/core-developers : Could you please, review this? Thanks! :)

strongbox-commons/src/main/resources/application.yaml
sbespalov: > Could we have this as a non-plain-text value? @strongbox/core-developers : Any suggestions?

we can try to research maybe this is possible to use certificate for this

> And... perhaps the application.yaml should only be viewable by the owner (we can control this in the strongbox-distribution's assembly descriptor).

AFAIK this `application.yaml` is under classpath, so you can't directly change or read it

strongbox-commons/src/main/resources/application.yaml
```diff
@@ -62,4 +62,31 @@ management:
...
+  enableMulticastConfig: true
```
sbespalov: @cbrianhill multicast should be disabled by default

sbespalov:
BTW, @carlspring 

should we expose all of this properties as env variables like we already have for [Application Properties](https://strongbox.github.io/user-guide/application-properties.html#application-properties)?

strongbox-commons/src/main/resources/application.yaml
cbrianhill: What do you think about `jasypt-spring-boot` for encrypting these properties (https://github.com/ulisesbocchio/jasypt-spring-boot)?

The `application.yaml` actually is baked into a jar file (the `strongbox-commons` JAR file which is then included in the spring-boot uber-jar.  Would it be a good convention for the uber-jar to be readable only by the owner?

strongbox-commons/src/main/resources/application.yaml
cbrianhill: Good catch, will disable this.  I'd flipped it for local testing.

cbrianhill:
> BTW, @carlspring
> 
> should we expose all of this properties as env variables like we already have for [Application Properties](https://strongbox.github.io/user-guide/application-properties.html#application-properties)?

Yeah, I considered this, please let me know your thoughts.  It would be quite a few environment variables, but happy to wire them up.

strongbox-commons/src/main/resources/application.yaml
carlspring: @cbrianhill ,

> The application.yaml actually is baked into a jar file (the strongbox-commons JAR file which is then included in the spring-boot uber-jar. Would it be a good convention for the uber-jar to be readable only by the owner?

We actually need to do something else... 

At the moment, all the properties are being loaded from the classpath. We should check, if they are actually available on the file system and load them from there first and only fallback to the classpath (from within the jar).

carlspring:
@sbespalov ,

> should we expose all of this properties as env variables like we already have for [Application Properties](https://strongbox.github.io/user-guide/application-properties.html#application-properties)?

If you're referring to the ones from the `application.yaml`, I don't think this would be a reasonable idea.

steve-todorov:
I agree with @carlspring - exposing this configuration as env vars would be a bit of an overkill and a possible security risk. The password is currently in plain-text and even if it weren't exposing the password as an env variable is probably not a very good idea.

sbespalov:
ok, thanks for clarifying
in general this PR seems very good, thanks @cbrianhill

carlspring:
Thanks for this fix, @cbrianhill ! :)

PR #1433, opened 2019-08-18
Issue-1430 java.lang.StackOverflowError on cancelled download request

### LINKED ISSUE #1430: java.lang.StackOverflowError on cancelled download request
# Bug Description

When we cancel file download, `java.lang.StackOverflowError` occurs in logs.

# Steps To Reproduce

0. Run `strongbox`
1. Hit http://localhost:48080/storages/storage-common-proxies/maven-central/.index/nexus-maven-repository-index.gz in the browser (wait for the file if it's locked)
2. Cancel a running download
3. See `strongbox` logs

# Expected Behavior

Proper handle file download cancel request.

# Logs

https://gist.github.com/fuss86/f5fab07385f9e1ac89976ff1e17fae9a

# Environment

* Strongbox Version: 
```
{"strongbox":{"version":"1.0-SNAPSHOT","revision":"868626f3ec2f0616"}}
```
* Maven version (please provide the full output of `mvn --version`):
```
$ mvn --version
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T21:39:06+02:00)
Maven home: C:\tools\apache-maven-3.5.0
Java version: 1.8.0_161, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk1.8.0_161\jre
Default locale: pl_PL, platform encoding: Cp1250
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
```

BogdanSukonnov:
Catch the EOFException with cause message `Connection reset by peer` in OutputStream.write() and flush(). Change request status to `202 ACCEPTED` for debug logs.

This pull request closes #1430 

# Acceptance Test

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes, please see #...
  * [x] No

  * [x] No


fuss86:
Thanks @BogdanSukonnov for your PR !

Could you please share with us a [gist ](https://gist.github.com/) containing logs/exception when you cancelled download request ?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java
```diff
@@ -297,8 +299,20 @@ protected void copyToResponse(InputStream is,
...
+                    os.flush();
+                }
+                catch (EOFException e)
+                {                   
+                    if (e.getCause().getMessage().equals("Connection reset by peer")) {
+                        logger.debug("Socket has been closed. Possibly, user cancelled download.");
+                        // Set response status 202 ACCEPTED just for logs
+                        response.setStatus(202);
+                        break;
+                    }
```
fuss86: otherwise we should re-throw the `EOFException` rather than do nothing

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java
```diff
@@ -297,8 +299,20 @@ protected void copyToResponse(InputStream is,
...
-                os.write(bytes, 0, readLength);
-                os.flush();
+                try
+                {
+                    os.write(bytes, 0, readLength);
+                    os.flush();
+                }
+                catch (EOFException e)
+                {                   
+                    if (e.getCause().getMessage().equals("Connection reset by peer")) {
```
fuss86: please also check if `IOException.class.equals(e.getCause().getClass())`, There are other possibilities that may cause `Connection reset by peer` message, see this thread for example: https://stackoverflow.com/questions/20741405/javax-net-ssl-sslexception-ssl-handshake-aborted-connection-reset-by-peer-while/20745624

BogdanSukonnov: Thanks, man! I really appreciate your help! My intention here was to catch  explicitly EOFException, which extends IOException. I was sure that all other exceptions can't be catches with this code. And only after that i check message. Do i really need to check class?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java
BogdanSukonnov: Thanks! I'll fix this.

sbespalov:
@fuss86 @BogdanSukonnov shouldn't this be placed in `DefaultExceptionHandler`?

BogdanSukonnov:
> shouldn't this be placed in `DefaultExceptionHandler` instead of `BaseController`?

I'll try to catch EOFException in DefaultExceptionHandler. Not sure  that I can distinguish OutputStream from InputStream there.

sbespalov:
@BogdanSukonnov never mind, seems you right and this should be in `BaseController`. Just checking options, thanks for clarifying :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java
BogdanSukonnov: Done

fuss86: `else` is unnecessary

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/BaseController.java
fuss86: Sorry, but I meant `e.getCause()`, not `e`.

BogdanSukonnov:
> Could you please share with us a [gist ](https://gist.github.com/) containing logs/exception when you cancelled download request ?

[logs after fix](https://gist.github.com/BogdanSukonnov/342676f17e0203eb0b8c1f25170b7770)

BogdanSukonnov:
@fuss86 can you help me to understand the state of this PR? Do you need to confirm that requestaed changes are done? Or it just waits for merging by team?

fuss86:
There is one unresolved comment. I think we should add a check `IOException.class == e.getCause().getClass()`.
@sbespalov what do you think ?

sbespalov:
@fuss86 good that you asked. I actually just noticed that in your [gist](https://gist.github.com/fuss86/f5fab07385f9e1ac89976ff1e17fae9a#file-201908142121-log-L19) there is `org.eclipse.jetty.io.EofException` but in `BaseController` the `java.io.EOFException` catches. @BogdanSukonnov  is there any reason why we can't catch original `org.eclipse.jetty.io.EofException`? It's seems that `org.eclipse.jetty.io.EofException` within flush should be enough and we even don't need to check the cause message.

The other thing is that we probably better decorate the `response.getOutputStream()` with something like `ProxyOutputStream` and `try...catch` the flush there to be sure that we have what expected within this issue. After this we can re-throw something custom like `ConnectionResetByPeerException` and catch this exception in `BaseController` in turn.

BogdanSukonnov:
@sbespalov I use `java.io.EOFException` as server independent solution. Didn't find any advantage in using `org.eclipse.jetty.io.EofException`. As said in docs 

> The only difference in handling is that Jetty EOFs are logged less verbosely.

 I believe that if we need to check message (do not sure about this either) in `BaseController` - we need to do it with both of this classes. Of course if we want to catch exception not in `BaseCotroller` than `org.eclipse.jetty.io.EofException` will help us to distinguish between EOF received from the connection from any other case.  But i'm a beginner, so you definetly know better. 
About this decorator pattern I'm afraid it's too much for me. You just will waist to much time, correcting my decisions. I don't know. What do you think?

carlspring:
@sbespalov , @fuss86 ,

@steve-todorov and I have been thinking of some day investigating Netty (for a few reasons I won't get into right now) and perhaps it would be better to keep this generic for now, unless you feel strongly.

carlspring:
@sbespalov / @fuss86 ,

I have reviewed this and I think it looks okay. Feel free to merge it when both of you are happy with the changes.

carlspring:
@BogdanSukonnov ,

Thank you very much for your help and congratulations on your first pull request!

sbespalov:
@BogdanSukonnov regarding the fact that `org.eclipse.jetty.io.EofException` is platform specific the thing is that this issue in general is platform specific and even with `java.io.EOFException` there is a chance that Tomcat or Netty will throw something else and this will not work like this, especially with message verification.

Purposed solution assumed to use exception handling logic within composition, and `OutputStream` decorator implementation can be different depending on environment.

Anyway, it's seems we have following options:
1. As most of the people happy with current implementation we can merge it as it is and get back to it in the future if needed.
2. If you interested to learn some patterns then we can continue and I will help to make it generic and flexible for real with decorated stream solution proposed above.

BogdanSukonnov:
@sbespalov it's really interesting, I'll try Tomcat and may be JetFly. And to practice some patterns sounds good to me. Thanks!

sbespalov:
@BogdanSukonnov ok, please proceed then

To be clear regarding Jetty specific decorator I mean that you should extend the `org.apache.commons.io.output.ProxyOutputStream` which will rethrow the `org.eclipse.jetty.io.EofException`  as `Http202PropogateException`. 

Also if need more help then I am available at our [chat](https://chat.carlspring.org/).

BogdanSukonnov:
@sbespalov please check the last commit. Is it what you ment?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/DefaultExceptionHandler.java
```diff
@@ -87,6 +88,11 @@
         HttpHeaders headers = new HttpHeaders();
         return handleExceptionInternal(ex, body, headers, HttpStatus.SERVICE_UNAVAILABLE, request);
     }
+    
+    @ExceptionHandler(Http202PropogateException.class)
+    protected void handleHttp202PropogateException(Exception ex) {
+        logger.debug(ex.getMessage());
```
sbespalov: @BogdanSukonnov you should provide HTTP 202 response code here as well.

strongbox-web-core/src/main/java/org/carlspring/strongbox/exception/ExceptionHandlindOutputStream.java
```diff
@@ -0,0 +1,39 @@
...
+    {
+        super(proxy);
+    }
+
+    @Override
+    protected void handleIOException(IOException e)
+        throws IOException
+    {
+        if (e.getClass().equals(EofException.class)) {
+            throw new Http202PropogateException("Socket has been closed. Possibly, user cancelled download.");
```
sbespalov: it should be better to set source `EofException` as `Http202PropogateException` cause.

sbespalov:
@BogdanSukonnov this looks good, just few comments.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/DefaultExceptionHandler.java
BogdanSukonnov: done

strongbox-web-core/src/main/java/org/carlspring/strongbox/exception/ExceptionHandlindOutputStream.java
BogdanSukonnov: done

strongbox-web-core/src/main/java/org/carlspring/strongbox/exception/ExceptionHandlindOutputStream.java
```diff
@@ -31,7 +28,7 @@ protected void handleIOException(IOException e)
         throws IOException
     {
         if (e.getClass().equals(EofException.class)) {
```
carlspring: Cosmetic: Braces on new line, please.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/DefaultExceptionHandler.java
```diff
@@ -90,8 +90,9 @@
     }
     
     @ExceptionHandler(Http202PropogateException.class)
-    protected void handleHttp202PropogateException(Exception ex) {
-        logger.debug(ex.getMessage());        
+    protected void handleHttp202PropogateException(Exception ex, HttpServletResponse httpResponse) {
```
carlspring: Cosmetic: Braces on new line, please.

carlspring:
@fuss86 , @sbespalov ,

Would you mind re-reviewing this? Thanks! :)

carlspring:
@BogdanSukonnov : Could you please rebase this pull request? Thanks! :)

carlspring:
@BogdanSukonnov ,

The username under which you're committing and pushing to Github seems to be mis-configured. You will have to look into how to configure your git settings properly, amend the author of the commits and then force push back to the branch. (When your user is properly set up, you see your avatar when against your commits in the commit history, otherwise it's a default grey avatar).

strongbox-web-core/src/main/java/org/carlspring/strongbox/exception/ExceptionHandlindOutputStream.java
BogdanSukonnov: sorry! old habits :) , fixed.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/DefaultExceptionHandler.java
BogdanSukonnov: fixed

BogdanSukonnov:
@carlspring thanks, Martin! The problem with user was just a dot in email in one place and no dot in another. I added both variants in GitHub profile. No need to force push.

carlspring:
@BogdanSukonnov : Thanks for fixing it! :)

strongbox-web-core/src/main/java/org/carlspring/strongbox/exception/ExceptionHandlindOutputStream.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import org.eclipse.jetty.io.EofException;
+import org.apache.commons.io.output.ProxyOutputStream;
+
+/**
+ * @author Bogdan Sukonnov
+ *
+ */
+public class ExceptionHandlindOutputStream extends ProxyOutputStream
```
carlspring: This class' name has a typo, it should be called `ExceptionHandlingOutputStream` (`handling`, not `handlind`).

Could you please correct this?

BogdanSukonnov: thanks, Martin! fixed it

carlspring: Thanks! :)

PR #1425, opened 2019-08-08
Issue-950: Make the strongbox-storage-api's tests work in parallel

ptirador:
# Pull Request Description

This pull request closes #950 

# Acceptance Test

* [x] Building the code with `mvn clean install -Dintegration.tests` still works.
* [x] Running `mvn spring-boot:run` in the `strongbox-web-core` still starts up the application correctly.
* [x] Building the code and running the `strongbox-distribution` from a `zip` or `tar.gz` still works.
* [x] The tests in the [`strongbox-web-integration-tests`](https://github.com/strongbox/strongbox-web-integration-tests/) still run properly.

# Questions

* Does this pull request break backward compatibility? 
  * [ ] Yes
  * [x] No

* Does this pull request require other pull requests to be merged first? 
  * [ ] Yes, please see #...
  * [x] No

* Does this require an update of the documentation?
  * [ ] Yes, please see strongbox/strongbox-docs#{PR_NUMBER}
  * [x] No

PR #1451, opened 2019-08-30
Revert "Issue-950: Make the strongbox-storage-api's tests work in parallel"

carlspring:
Reverts strongbox/strongbox#1425


steve-todorov:
This is no longer necessary in favor of #1454.

PR #1520, opened 2019-10-12
Issue 1513: Unit test for RemoteRepositoryAlivenessCacheManager

mwvdev:
- Added initial test using a mocked remote repository.
- The putShouldChangeCachedValueImmediately parameterized test fails in two cases, which merits further investigation. This might be a problem with the Hazelcast Cache not updating values immediately.
- Eviction strategy for isAliveShouldReturnTrueWhenRemoteRepositoryCachedValueIsExpired still to be done.

# Pull Request Description

This pull request is a draft. I'd appreciate a bit of feedback on this as-is.

I would like someone to take a look at the parameterized test that currently fails. It seems to me this should be passing, so perhaps there's an issue with the Hazelcast cache.

A bit of help on how to forcibly expire a cache item would be appreciated. I could forcibly just evict the cache entry by using `Cache.evict`, but could be a better way.

# Acceptance Test

# Questions

  * [ ] Yes
  * [X] No

  * [ ] Yes, please see #...
  * [X] No

  * [X] No


carlspring:
@mwvdev ,

Thanks for your pull request!

Would you mind adding some more details, explaining the error and adding the exception you're getting? Thanks! :)

@sbespalov , @fuss86 : Could you please have a look and review? Thanks! :)

mwvdev:
Actually, seems I was a bit too fast on the trigger. I've found a few mistakes that I'm going to be looking into fixing. The `isAliveShouldReturnFalseWhenRemoteRepositoryCachedValueIsFalse` is not asserting what it claims to be... right now it's just asserting true, which indicates there's probably a more general problem me assigning values to the cache. 

Having looked at it, I'm actually not able to assign any values to the cache, it seems to just rely on the fact that `isAlive` returns `true` for when the cache either contains a `null` or `true` value.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@SpringBootTest
+@ActiveProfiles({ "test",
+                  "RemoteRepositoryAlivenessCacheManagerTestConfig" })
+@TestExecutionListeners(listeners = { CacheManagerTestExecutionListener.class },
+                        mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
+public class RemoteRepositoryAlivenessCacheManagerTest
```
fuss86: `strongbox-storage-api` tests are configured to be run `concurrently` (see `src/test/resources/junit-platform.properties`). You should make sure every method uses its own `remoteRepository` or you should mark the test as `@Execution(SAME_THREAD)`. Otherwise the results are unpredictable.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,140 @@
...
+    {
+        initializeCache(false);
+
+        assertTrue(remoteRepositoryAlivenessCacheManager.isAlive(remoteRepository));
+    }
+
+    @Test
+    public void isAliveShouldReturnTrueWhenRemoteRepositoryCachedValueIsExpired()
+    {
+        // TODO: Figure out how to forcibly expire a cached item.
```
fuss86: consider `com.hazelcast.config.MapConfig#setTimeToLiveSeconds`

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Hi, thanks for your feedback. I have inlined the instances of `remoteRepository`, please have a look at the latest commits.

mwvdev:
@fuss86 I'm having troubles getting anything to actually be persisted in the cache. I've tried debugging through the various steps but I'm not sure why `put` operations aren't persisted in the cache. Could you perhaps have a look at the `isAliveShouldReturnFalseWhenRemoteRepositoryCachedValueIsFalse` unit test to see if you can spot any issues?

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    private void initializeCache(Boolean initialValue)
+    {
+        Cache cache = cacheManager.getCache(CacheName.Repository.REMOTE_REPOSITORY_ALIVENESS);
+        Objects.requireNonNull(cache, "remoteRepositoryAliveness cache configuration was not provided");
+
+        cache.put(REMOTE_REPOSITORY_URL, initialValue);
+    }
+
+    @Profile("RemoteRepositoryAlivenessCacheManagerTestConfig")
+    @Import({ HazelcastConfiguration.class,
```
sbespalov: @mwvdev probably you only need `HazelcastConfiguration` here, below configurations seems excess.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
sbespalov: it seems we better use `@Execution(SAME_THREAD)` here because test methods will use same cache instance anyway.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void isAliveShouldReturnTrueWhenRemoteRepositoryNotCached()
+    {
+        RemoteRepository remoteRepository = createMockedRemoteRepository();
+
+        boolean alive = remoteRepositoryAlivenessCacheManager.isAlive(remoteRepository);
+
+        assertTrue(alive);
```
sbespalov: in latter days we use `org.assertj` :)

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void isAliveShouldReturnFalseWhenRemoteRepositoryCachedValueIsFalse()
+    {
+        RemoteRepository remoteRepository = createMockedRemoteRepository();
+
+        initializeCache(false);
```
sbespalov: seems this `initializeCache` is excess because of below  `remoteRepositoryAlivenessCacheManager.put`

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideCacheValues")
+    public void putShouldChangeCachedValueImmediately(Boolean initialCacheValue,
+                                                      boolean newCacheValue)
+    {
+        RemoteRepository remoteRepository = createMockedRemoteRepository();
+
+        initializeCache(initialCacheValue);
```
sbespalov: seems this `initializeCache` is excess because of below  `remoteRepositoryAlivenessCacheManager.put`

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        remoteRepositoryAlivenessCacheManager = new RemoteRepositoryAlivenessCacheManager(cacheManager);
+    }
+
+    @Test
+    public void isAliveShouldReturnTrueWhenRemoteRepositoryNotCached()
+    {
+        RemoteRepository remoteRepository = createMockedRemoteRepository();
+
+        boolean alive = remoteRepositoryAlivenessCacheManager.isAlive(remoteRepository);
```
sbespalov: before this we probably should check that there is no cached value in `Cache`

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
sbespalov: you can create just `CahceManager` like below:
```
    @Bean
    public CacheManager cacheManager(HazelcastInstance hazelcastInstance)
    {
        return new HazelcastCacheManager(hazelcastInstance);
    }
```

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Test
+    public void isAliveShouldReturnTrueWhenRemoteRepositoryCachedValueIsExpired()
+    {
+        // TODO: Figure out how to forcibly expire a cached item.
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideCacheValues")
+    public void putShouldChangeCachedValueImmediately(Boolean initialCacheValue,
```
sbespalov: this parameterized method looks very nice :+1:

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Great suggestion, I'll use this approach instead.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: I was debating a bit whether to just use `remoteRepositoryAlivenessCacheManager.put` but decided against it to avoid risking the possibility of a bug in `remoteRepositoryAlivenessCacheManager.put` affecting the other tests. We're going to be testing `remoteRepositoryAlivenessCacheManager.put` so that's not a valid concern. I'll remove `initializeCache`.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: I agree, never hurts to verify assumptions.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Thank you :-)

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Ah, nice catch. I agree, I'll revert the previous fix and add `@Execution(SAME_THREAD)` instead.

mwvdev:
> @mwvdev thanks for the PR!
> 
> can you please proceed with comments?

@sbespalov Thanks for the feedback! I'll start working my way through the comments.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Roger that, changed.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: Having looked this over again I'm actually leaning toward keeping `initializeCache` and removing the call to `remoteRepositoryAlivenessCacheManager.put` in this test.

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/storage/repository/remote/heartbeat/RemoteRepositoryAlivenessCacheManagerTest.java
mwvdev: I added this to ensure that `remoteRepositoryAlivenessCacheManager.put` can immediately change the value from akk the possible initial states: `null`, `false`, `true`. This is basically to ensure there's not a corner-case when switching from i.e. a `null` to `true`. If you haven't been persuaded by arguments, then just let me know and I'll refactor it :-)

PR #1577, opened 2019-11-08
Implemented strict checksum validation

### LINKED ISSUE #1512: Implement strict checksum validation option for Repository
# Task Description
We need to have an  `strictChecksum` option for `Repository` which will prevent of uploading artifacts with broken checksum. For now we have only `ERROR` log message in case of invalid checksum _(see `ArtifactManagementService.validateUploadedChecksumAgainstCache`)_. 

This should be configurable.

# Tasks

* [ ] Add `strictChecksum` property of type `boolean` for `org.carlspring.strongbox.storage.repository.Repository` interface
* [] Support `strictChecksum` option for `repositories` items in `strongbox.yaml`
* [ ] Throw `ArtifactStorageException` in case of `matchesChecksum` returns `false` and `Repository.isStrictChecksum() == true` within `ArtifactManagementService#validateUploadedChecksumAgainstCache`
* [ ] Provide Unit Tests covering new feature

# Useful Links

* [strongbox.yaml](https://strongbox.github.io/user-guide/configuration-files/strongbox-yaml.html)
* [Writing tests](https://strongbox.github.io/developer-guide/writing-tests.html)

# Help

* Points of contact:
  * @carlspring
  * @fuss86
  * @sbespalov
  * @steve-todorov

newarcher36:
# Pull Request Description

This pull request closes https://github.com/strongbox/strongbox/issues/1512

# Acceptance Test

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes, please see #...
  * [x] No

  * [x] Yes, please see strongbox/strongbox-docs#53
  * [ ] No

### DIFF
--- a/strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.yaml
+++ b/strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.yaml
@@ -51,6 +51,7 @@ configuration:
           type: hosted
           allowsForceDeletion: true
           checksumHeadersEnabled: true
+          strictChecksumValidation: true
           repositoryConfiguration:
             type: Maven 2
             indexingEnabled: true
@@ -63,6 +64,7 @@ configuration:
           type: hosted
           secured: true
           checksumHeadersEnabled: true
+          strictChecksumValidation: true
           repositoryConfiguration:
             type: Maven 2
             indexingEnabled: true
@@ -286,6 +288,7 @@ configuration:
           type: hosted
           allowsForceDeletion: true
           checksumHeadersEnabled: true
+          strictChecksumValidation: true
         nuget-third-party:
           id: nuget-third-party
           policy: release
--- a/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
+++ b/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
@@ -117,6 +117,7 @@ public class ArtifactManagementService
             logger.debug("Stored [{}] bytes for [{}].", result, repositoryPath);
             aos.flush();
         }
+
         catch (IOException e)
         {
            throw e; 
@@ -230,7 +231,7 @@ public class ArtifactManagementService
                          artifactPath,
                          checksumStr);
 
-            if ("Maven 2".equals(repository.getLayout()) && repository.isStrictChecksumValidation())
+            if (repository.isStrictChecksumValidation())
             {
                 throw new ArtifactStorageException(String.format("Invalid checksum [%s] for artifact [%s]", checksumStr, artifactPath));
             }
--- a/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
+++ b/strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
@@ -184,7 +184,7 @@ public class ArtifactManagementService
             if (checksumValue != null && checksumValue.length > 0)
             {
                 // Validate checksum with artifact digest cache.
-                validateUploadedChecksumAgainstCache(checksumValue, repositoryPathId);
+                validateUploadedChecksumAgainstCache(checksumValue, repositoryPathId, repository);
             }
         }
         
@@ -212,7 +212,9 @@ public class ArtifactManagementService
     }
 
     private void validateUploadedChecksumAgainstCache(byte[] checksum,
-                                                      URI artifactPathId)
+                                                      URI artifactPathId,
+                                                      Repository repository)
+            throws ArtifactStorageException
     {
         logger.debug("Received checksum: {}", new String(checksum, StandardCharsets.UTF_8));
 
@@ -222,12 +224,19 @@ public class ArtifactManagementService
 
         if (!matchesChecksum(checksum, artifactBasePath, checksumExtension))
         {
+            String checksumStr = new String(checksum, StandardCharsets.UTF_8);
+
             logger.error("The checksum for {} [{}] is invalid!",
                          artifactPath,
-                         new String(checksum, StandardCharsets.UTF_8));
-        }
+                         checksumStr);
 
-        checksumCacheManager.removeArtifactChecksum(artifactBasePath, checksumExtension);
+            if ("Maven 2".equals(repository.getLayout()) && repository.isStrictChecksumValidation())
+            {
+                throw new ArtifactStorageException(String.format("Invalid checksum [%s] for artifact [%s]", checksumStr, artifactPath));
+            }
+
+            checksumCacheManager.removeArtifactChecksum(artifactBasePath, checksumExtension);
+        }
     }
 
     private boolean matchesChecksum(byte[] pChecksum,
--- a/strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/storage/repository/RepositoryAttributes.java
+++ b/strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/storage/repository/RepositoryAttributes.java
@@ -40,4 +40,6 @@ public @interface RepositoryAttributes
 
     boolean checksumHeadersEnabled() default false;
 
+    boolean strictChecksumValidation() default false;
+
 }
--- a/strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryDto.java
+++ b/strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryDto.java
@@ -58,6 +58,8 @@ public class RepositoryDto
 
     private boolean checksumHeadersEnabled;
 
+    private boolean strictChecksumValidation;
+
     /**
      * The per-repository proxy settings that override the overall global proxy settings.
      */
@@ -223,9 +225,20 @@ public class RepositoryDto
         return checksumHeadersEnabled;
     }
 
+    public boolean isStrictChecksumValidation()
+    {
+        return strictChecksumValidation;
+    }
+
     public void setChecksumHeadersEnabled(boolean checksumHeadersEnabled)
     {
         this.checksumHeadersEnabled = checksumHeadersEnabled;
+    }
+
+    public void setStrictChecksumValidation(boolean strictChecksumValidation)
+    {
+
+        this.strictChecksumValidation = strictChecksumValidation;
     }
 
     public MutableProxyConfiguration getProxyConfiguration()
--- a/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
+++ b/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
@@ -5,6 +5,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import org.carlspring.strongbox.artifact.ArtifactTag;
 import org.carlspring.strongbox.artifact.MavenArtifact;
 import org.carlspring.strongbox.artifact.MavenArtifactUtils;
+import org.carlspring.strongbox.artifact.generator.ArtifactGenerator;
+import org.carlspring.strongbox.artifact.generator.MavenArtifactGenerator;
 import org.carlspring.strongbox.config.Maven2LayoutProviderTestConfig;
 import org.carlspring.strongbox.domain.ArtifactEntry;
 import org.carlspring.strongbox.providers.ProviderImplementationException;
@@ -15,6 +17,7 @@ import org.carlspring.strongbox.providers.io.RepositoryPathResolver;
 import org.carlspring.strongbox.providers.io.RepositoryStreamSupport.RepositoryInputStream;
 import org.carlspring.strongbox.repository.MavenRepositoryFeatures;
 import org.carlspring.strongbox.storage.ArtifactStorageException;
+import org.carlspring.strongbox.storage.checksum.ChecksumCacheManager;
 import org.carlspring.strongbox.storage.metadata.MavenSnapshotManager;
 import org.carlspring.strongbox.storage.repository.Repository;
 import org.carlspring.strongbox.storage.repository.RepositoryPolicyEnum;
@@ -31,6 +34,7 @@ import javax.inject.Inject;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
@@ -48,6 +52,7 @@ import java.util.stream.Stream;
 
 import org.apache.commons.io.FilenameUtils;
 import org.apache.maven.artifact.Artifact;
+import org.checkerframework.checker.units.qual.C;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.parallel.Execution;
@@ -114,6 +119,9 @@ public class ArtifactManagementServiceImplTest
 
     @Inject
     private RepositoryPathResolver repositoryPathResolver;
+
+    @Inject
+    protected ChecksumCacheManager checksumCacheManager;
 
     @ExtendWith({ RepositoryManagementTestExecutionListener.class,
                   ArtifactManagementTestExecutionListener.class })
@@ -690,27 +698,24 @@ public class ArtifactManagementServiceImplTest
     @Test
     public void testFailArtifactStorageWhenChecksumIsBroken(@MavenRepository(repositoryId = "broken-checksum-test")
                                                             @RepositoryAttributes(strictChecksumValidation = true)
-                                                            Repository repository,
-                                                            @MavenTestArtifact(repositoryId = "broken-checksum-test",
-                                                                               id = "org.carlspring.strongbox:strongbox-test-artifact",
-                                                                               versions = { "1.0" })
-                                                            Path artifactPath)
+                                                                    Repository repository,
+                                                            @MavenTestArtifact(
+                                                                    id = "org.carlspring.strongbox:strongbox-test-artifact",
+                                                                    versions = { "1.0" })
+                                                                    Path artifactPath)
             throws IOException
     {
-        final String storageId = repository.getStorage().getId();
-        final String repositoryId = repository.getId();
         final String checksumFilePath = "org/carlspring/strongbox/strongbox-test-artifact/1.0/strongbox-test-artifact-1.0.jar.md5";
 
-        RepositoryPath repositoryPath = repositoryPathResolver.resolve(storageId,
-                                                                       repositoryId,
-                                                                       checksumFilePath);
-        Files.write(repositoryPath, "broken_checksum".getBytes());
+        RepositoryPath repositoryPath = repositoryPathResolver.resolve(repository).resolve(checksumFilePath);
+        String artifactBasePath = repositoryPath.toUri().toString();
+        checksumCacheManager.addArtifactChecksum(artifactBasePath, "MD5", "XYZ");
 
         try (InputStream is = Files.newInputStream(repositoryPath))
         {
             Throwable throwable = catchThrowable(() -> mavenArtifactManagementService.store(repositoryPath,is));
 
-            final String fullChecksumFilePath = String.format("/%s/%s/%s", storageId, repositoryId, checksumFilePath);
+            final String fullChecksumFilePath = "";//String.format("/%s/%s/%s", storageId, repositoryId, checksumFilePath);
             assertThat(throwable)
                     .isExactlyInstanceOf(ArtifactStorageException.class)
                     .hasMessage("Invalid checksum [broken_checksum] for artifact [strongbox:%s]",
--- a/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
+++ b/strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
@@ -7,6 +7,8 @@ import org.carlspring.strongbox.artifact.MavenArtifact;
 import org.carlspring.strongbox.artifact.MavenArtifactUtils;
 import org.carlspring.strongbox.config.Maven2LayoutProviderTestConfig;
 import org.carlspring.strongbox.domain.ArtifactEntry;
+import org.carlspring.strongbox.providers.ProviderImplementationException;
+import org.carlspring.strongbox.providers.io.RepositoryFileAttributeType;
 import org.carlspring.strongbox.providers.io.RepositoryFiles;
 import org.carlspring.strongbox.providers.io.RepositoryPath;
 import org.carlspring.strongbox.providers.io.RepositoryPathResolver;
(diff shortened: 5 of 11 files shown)


strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryDto.java
```diff
@@ -223,11 +225,21 @@ public boolean isChecksumHeadersEnabled()
...
+    {
+        return strictChecksum;
+    }
+
     public void setChecksumHeadersEnabled(boolean checksumHeadersEnabled)
     {
         this.checksumHeadersEnabled = checksumHeadersEnabled;
     }
 
+    public void setStrictChecksum(boolean strictChecksum) {
```
carlspring: Cosmetic: Braces on new lines please. :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
```diff
@@ -679,6 +685,36 @@ public void testChecksumsStorage(@MavenRepository(repositoryId = "checksums-stor
...
+    @ExtendWith({ RepositoryManagementTestExecutionListener.class,
+                  ArtifactManagementTestExecutionListener.class })
+    @Test
+    public void testFailArtifactStorageWhenChecksumIsBroken(@MavenRepository(repositoryId = "broken-checksum-test")
+                                                            @RepositoryAttributes(strictChecksum = true)
+                                                            Repository repository,
+                                                            @MavenTestArtifact(repositoryId = "broken-checksum-test",
+                                                                    id = "org.carlspring.strongbox:strongbox-test-artifact",
+                                                                    versions = { "1.0" })
+                                                            Path artifactPath)
```
carlspring: Cosmetic: Please, re-indent this:
```
    @Test
    public void testFailArtifactStorageWhenChecksumIsBroken(@MavenRepository(repositoryId = "broken-checksum-test")
                                                            @RepositoryAttributes(strictChecksum = true)
                                                            Repository repository,
                                                            @MavenTestArtifact(repositoryId = "broken-checksum-test",
                                                                               id = "org.carlspring.strongbox:strongbox-test-artifact",
                                                                               versions = { "1.0" })
                                                            Path artifactPath)
```

strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.yaml
```diff
@@ -51,6 +51,7 @@ configuration:
           type: hosted
           allowsForceDeletion: true
           checksumHeadersEnabled: true
+          strictChecksum: true
```
carlspring: I think this should be called either `strictChecksumValidation`, or at the very least `strictChecksums`.

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -222,12 +224,19 @@ private void validateUploadedChecksumAgainstCache(byte[] checksum,
...
+
             logger.error("The checksum for {} [{}] is invalid!",
                          artifactPath,
-                         new String(checksum, StandardCharsets.UTF_8));
-        }
+                         checksumStr);
+
+            if (repository.isStrictChecksum())
+            {
+                throw new ArtifactStorageException(String.format("Checksum broken [%s] for artifact [%s]", checksumStr, artifactPath));
```
carlspring: Could you please change this to:
```
                throw new ArtifactStorageException(String.format("Invalid checksum [%s] for artifact [%s]", checksumStr, artifactPath));
```
Thanks! :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
```diff
@@ -679,6 +685,36 @@ public void testChecksumsStorage(@MavenRepository(repositoryId = "checksums-stor
...
+                                                                       checksumFilePath);
+        Files.write(repositoryPath, "broken_checksum".getBytes());
+
+        try (InputStream is = Files.newInputStream(repositoryPath))
+        {
+            Throwable throwable = catchThrowable(() -> mavenArtifactManagementService.store(repositoryPath,is));
+
+            assertThat(throwable)
+                    .isExactlyInstanceOf(ArtifactStorageException.class)
+                    .hasMessage("Checksum broken [broken_checksum] for artifact [strongbox:/storage0/broken-checksum-test/org/carlspring/strongbox/strongbox-test-artifact/1.0/strongbox-test-artifact-1.0.jar.md5]");
```
ptirador: Please, use this variable instead:
```
            final String fullChecksumFilePath = String.format("/%s/%s/%s", storageId, repositoryId, checksumFilePath);
            assertThat(throwable)
                    .isExactlyInstanceOf(ArtifactStorageException.class)
                    .hasMessage("Checksum broken [broken_checksum] for artifact [strongbox:" + fullChecksumFilePath + "]");
```

ptirador (changes requested):

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
Stephan202: AssertJ supports format strings,  so that last line can be:
```
.hasMessage("Checksum broken [broken_checksum] for artifact [strongbox:%s]", fullChecksumFilePath);
```
... which arguably reads nicer.

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/storage/repository/RepositoryDto.java
newarcher36: Done!

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
newarcher36: Done!

strongbox-resources/strongbox-storage-api-resources/src/main/resources/etc/conf/strongbox.yaml
newarcher36: Done!

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
newarcher36: Done!

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
newarcher36: Thank guys for your suggestions. Suggestions already aplplied

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
carlspring: Thanks! :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
```diff
@@ -679,6 +685,37 @@ public void testChecksumsStorage(@MavenRepository(repositoryId = "checksums-stor
...
+        Files.write(repositoryPath, "broken_checksum".getBytes());
+
+        try (InputStream is = Files.newInputStream(repositoryPath))
+        {
+            Throwable throwable = catchThrowable(() -> mavenArtifactManagementService.store(repositoryPath,is));
+
+            final String fullChecksumFilePath = String.format("/%s/%s/%s", storageId, repositoryId, checksumFilePath);
+            assertThat(throwable)
+                    .isExactlyInstanceOf(ArtifactStorageException.class)
+                    .hasMessage("Invalid checksum [broken_checksum] for artifact [strongbox:%s]", fullChecksumFilePath);
```
carlspring: Cosmetic: Would you mind re-aligning this code like this:
```
            assertThat(throwable).isExactlyInstanceOf(ArtifactStorageException.class)
                                 .hasMessage("Invalid checksum [broken_checksum] for artifact [strongbox:%s]",
                                             fullChecksumFilePath);
```

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/storage/repository/RepositoryAttributes.java
```diff
@@ -40,4 +40,6 @@
 
     boolean checksumHeadersEnabled() default false;
 
+    boolean strictChecksum() default false;
```
carlspring: Could you please change this as well, like we discussed earlier to: `strictChecksumValidation`? (And could you please check, if there are any other places like that need to be changed?). Thanks! :)

carlspring (changes requested):
@newarcher36 ,

I have some remarks:
* Could you please go over the `strongbox.yaml` and remove the `strictChecksumValidation: true` for all non-Maven repositories?
* Could you make this functionality the default behaviour only for Maven repositories?
* Could you please update the documentation in your other pull request
  * Regarding the property name in the `strongbox.yaml`
  * Mention that this property is switched on by default only for Maven repositories

Thanks for your help! :)

Martin

newarcher36:
Sure I am on it!

carlspring:
Great, thanks!

Could you also, please rebase? Thanks! :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
newarcher36: Done!

strongbox-storage/strongbox-storage-api/src/test/java/org/carlspring/strongbox/testing/storage/repository/RepositoryAttributes.java
newarcher36: Done!

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
```diff
@@ -222,12 +224,19 @@ private void validateUploadedChecksumAgainstCache(byte[] checksum,
...
         {
+            String checksumStr = new String(checksum, StandardCharsets.UTF_8);
+
             logger.error("The checksum for {} [{}] is invalid!",
                          artifactPath,
-                         new String(checksum, StandardCharsets.UTF_8));
-        }
+                         checksumStr);
+
+            if ("Maven 2".equals(repository.getLayout()) && repository.isStrictChecksumValidation())
```
carlspring: Please, use `Maven2LayoutProvider.ALIAS` for this.
Also, is it possible to move this check somewhere within the Maven layout provider's module's code? This service should be generic.

newarcher36: I have been searching for a place in Maven layout provider's module to perform "strictChecksumValidation". I found class ArtifactMetadataServiceImpl but not sure whether it is the best option. Any idea or inspiration that you could suggest guys? I am bit lost.
Thanks!

carlspring: @ptirador : Any advice?

ptirador: In my opinion, this could be done like this:
1. Replace this condition by a line of code like this:
```
validateChecksumForLayoutProvider(repository, checksumStr, artifactPath);
```

1.1 This method would be defined like this:
```
protected void validateChecksumForLayoutProvider(Repository repository,
                                                 String checksumStr,
                                                 String artifactPath)
            throws ArtifactStorageException
{
        // Empty for this generic class.
}
```

2. Create a class `org.carlspring.strongbox.services.impl. MavenArtifactManagementServiceImpl` in `strongbox-storage-maven-layout-provider`, which would extends from `ArtifactManagementService`.
2.1. Create a method `validateChecksumForLayoutProvider` in this class, which will override the method in `ArtifactManagementService`, adding this specific condition.

3. Search the uses of `ArtifactManagementService` in Maven-related classes (`MavenArtifactController`, etc) and replace them for the specific component `MavenArtifactManagementServiceImpl`.

Another option would be similar to this, but overriding the whole `validateUploadedChecksumAgainstCache` method instead of creating that method that I mentioned.

Yeah, maybe it's not the best solution (specially because of the last step), but I can't think of another solution right now...

sbespalov: > Also, is it possible to move this check somewhere within the Maven layout provider's module's code? This service should be generic.

this is not assumed to be layout specific. If it really should then we need a followup taks for it.

@carlspring can you please provide an example why you think it should depends on layout?

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
```diff
@@ -679,6 +685,38 @@ public void testChecksumsStorage(@MavenRepository(repositoryId = "checksums-stor
         assertThat(actualChecksums).isEqualTo(expectedChecksums);
     }
 
+    @ExtendWith({ RepositoryManagementTestExecutionListener.class,
+                  ArtifactManagementTestExecutionListener.class })
+    @Test
+    public void testFailArtifactStorageWhenChecksumIsBroken(@MavenRepository(repositoryId = "broken-checksum-test")
```
sbespalov: @newarcher36 the test case seems not really clear to me, can you please explain what you assumed here?

sbespalov (changes requested):
@newarcher36 thanks for the PR. Can you please proceed with test related comments?

@carlspring can you please also share your thoughts about layout specific checksum validation.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
newarcher36: Hi @sbespalov ,  I am assuming that the checksum of strongbox-test-artifact-1.0.jar.md5 differs from the checksum of the recent downloaded artifact. 

For instance, checksum of strongbox-test-artifact-1.0.jar.md5 is "broken_checksum" and the checksum of the downloaded artifact is "10d759d4a8a297bf5e28a6fc7eaf4b969d78a098".
I that case when is a Maven 2 layout, log and throw and ArtifactStorageException.

sbespalov: it seems currently it just checks that there was no checksums cached before checksum file stored, and there is also no point of [this](https://github.com/strongbox/strongbox/pull/1577/files#diff-a0ca838f38a43fbc4b2b2c83bf4b1fb5R707) write.

carlspring:
> @carlspring can you please also share your thoughts about layout specific checksum validation.

@sbespalov : I am not saying the the checksum validation should be layout-specific, I'm saying that it should not be enabled by default for all layout providers. npmjs.org and nuget.org don't have checksums associated with the artifacts that they host, do they? How does the checksum validation for these currently work, (if at all)? Are we parsing the metadata to get them? As far as I'm aware, we're not, but please feel free to correct me.

My point here is that if we enable this by default for all repositories and all layout providers, we'll probably break a lot of stuff. Will artifact downloads for proxy/group repositories work? I don't think we have a proper test case for this in place, ( @newarcher36 : perhaps you could put one together) ?

@sbespalov : I'm really looking forward to hearing your thoughts about this! :)

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
sbespalov: the main test case should be like follows:
1. you should not use `repositoryId` for `MavenTestArtifact`, because artifact should be deployed within test case.
2. you should store the artifact first to chache the checksum in `ChecksumCacheManager`
3. you should store the invalid checksum file and receive the exception

there is also should be a test case like you currenty have but without `Files.write`

and final test case should check valid case.

sbespalov:
@carlspring 
> npmjs.org and nuget.org don't have checksums associated with the artifacts that they host, do they?

they have checksums

>  How does the checksum validation for these currently work, (if at all)?

currntly it should work same way for all layouts

>Will artifact downloads for proxy/group repositories work? I don't think we have a proper test case for this in place, ( @newarcher36 : perhaps you could put one together) ?

we should probably have a test cases within this task as well

sbespalov:
also the layout specific test cases should be for controllers

carlspring:
> @carlspring
> 
> > npmjs.org and nuget.org don't have checksums associated with the artifacts that they host, do they?
> 
> they have checksums
> 

Right, I was under the impression that they didn't have them as files and/or that we weren't checking them for layout providers other than Maven. If you're saying that this is not the case, then could you please outline the sub-tasks that need to be done to check these as well? Thanks! :)

I think we should have a test per layout provider that does this:
* [ ] Check that strict checksum validation works for artifacts for this provider
* [ ] Check that strict checksum validation works for artifacts for in proxy repositories for this provider
* [ ] Check that strict checksum validation works for artifacts for in group repositories for this provider
* [ ] Log a warning, if strict checksum validation is disabled for the repository, but an artifact has been found to have a checksum mismatch.

Would this be enough?

> > How does the checksum validation for these currently work, (if at all)?
> 
> currntly it should work same way for all layouts
> 
> > Will artifact downloads for proxy/group repositories work? I don't think we have a proper test case for this in place, ( @newarcher36 : perhaps you could put one together) ?
> 
> we should probably have a test cases within this task as well

strongbox-storage/strongbox-storage-api/src/main/java/org/carlspring/strongbox/services/ArtifactManagementService.java
newarcher36: > **this** is not assumed to be layout specific. If it really should then we need a follow up tasks, but I'm not sure we need it now.

@sbespalov what do you mean about "this", you mean "strict checksum validation", so "strict checksum validation" should be generic for all layouts?

newarcher36:
Please correct me if I am mistaken:

1. Strict checksum validation is supposed to be generic for all layouts but only enabled by default  for Maven layout and disabled for the rest of layouts.

2. Provide with the following test cases :
 - Check that strict checksum validation works for artifacts for this provider.
 - Check that strict checksum validation works for artifacts in proxy repositories for this provider.
 - Check that strict checksum validation works for artifacts in group repositories for this provider.

So perform all tests mentioned above for each provider Maven, NuGet, Npm?

Thanks!

sbespalov:
> Strict checksum validation is supposed to be enabled by default for all layouts providers at the end? Which behaviour are we gonna follow?

I vote for it to be for all layouts. @carlspring please confirm.

> I am supposed to provide with the following test cases for each layout provider as well?

We should have such test cases for all ArtifactControllers (see `BaseArtifactController`). Also there probably no point to test this for group and proxy repositories as they assumed to be read only. 
The last thing is that we should have the test case for `ArtifactManagementService`, like already done here, but it should be generic with `NullLayoutProvider` instead of Maven related.

newarcher36:
@sbespalov All right, then I am gonna start with digging on how to perform test cases for all ArtifactControllers and check out BaseController.

carlspring:
> > Strict checksum validation is supposed to be enabled by default for all layouts providers at the end? Which behaviour are we gonna follow?
> 
> I vote for it to be for all layouts. @carlspring please confirm.

Very well, let's go with this option and see, if it breaks anything.

> > I am supposed to provide with the following test cases for each layout provider as well?
> 
> We should have such test cases for all ArtifactControllers (see `BaseArtifactController`). Also there probably no point to test this for group and proxy repositories as they assumed to be read only.

I agree.

> The last thing is that we should have the test case for `ArtifactManagementService`, like already done here, but it should be generic with `NullLayoutProvider` instead of Maven related.

+1.

carlspring:
@newarcher36 ,

Please, feel free to proceed! Thanks! :)

newarcher36:
Hi guys I am experiencing some troubles with the following test :

ArtifactManagementServiceImplTest#testFailArtifactStorageWhenChecksumIsBroken

I am getting an ArtifactNotFoundException but do not know why. I have been spending somte time but cannot figure out. Could you please have a look whenever you can?

Regards

ptirador:
Hi @newarcher36, I've been trying to reproduce the issue that you're having but I'm not able to. Coudl you rebase your branch and try it again?

newarcher36:
@ptirador I could solve it but thank you very much for checking out!

steve-todorov:
hey guys, what's the state of this pull request? @newarcher36 would you mind `rebasing` please?

// cc @sbespalov , @carlspring , @ptirador

newarcher36:
> hey guys, what's the state of this pull request? @newarcher36 would you mind `rebasing` please?
> 
> // cc @sbespalov , @carlspring , @ptirador

Yes I am still working on it, It is already rebased.

strongbox-storage/strongbox-storage-layout-providers/strongbox-storage-maven-layout/strongbox-storage-maven-layout-provider/src/test/java/org/carlspring/strongbox/services/ArtifactManagementServiceImplTest.java
newarcher36: Hi @sbespalov, thank you very much for providing your help I already fixed the test following your instructions but I still need to use Files.write in order to write and invalid checksum value in the md5 file and then store this invalid checksum md5 file in order to trigger ArtifactStorageException otherwise I do not know how can I create a file with an invalid checksum...

carlspring:
@newarcher36 ,

What is the status of this pull request?

steve-todorov:
@newarcher36 - it has been more than 5 months since you've started working on this. Will you be able to complete it any time soon? There are people who could pick it up from where you left and finalize it if you're too busy at work. :)

PR #1607, opened 2019-11-21
Reduce irrelevant Debug logs

anki2189:
# Pull Request Description

This pull request closes #

# Acceptance Test

# Questions

  * [ ] Yes
  * [ ] No

  * [ ] Yes, please see #...
  * [ ] No

  * [ ] No


anki2189:
@carlspring Pls review the PR.

anki2189:
> First of all, almost all of our modules are configured to produce a `test-jar` (containing their code from `${basedir}/test/classes` (this also included the resource files from `${basedir}/src/test/resources`)). This is done by adding the following to your `pom.xml`:
> 
> ```
>             <plugin>
>                 <groupId>org.apache.maven.plugins</groupId>
>                 <artifactId>maven-jar-plugin</artifactId>
>                 <executions>
>                     <execution>
>                         <goals>
>                             <goal>test-jar</goal>
>                         </goals>
>                     </execution>
>                 </executions>
>             </plugin>
> ```
> 
> (I can see that this has not been added to the `strongbox-common-resources` module, which is most-likely, why it didn't work for you, as you told me the other day in the chat).
> It was miss from my side as I haven't explored how this is working in other modules. I will try and test this again with suggested changes. 
> Secondly, if we need a `logback-test.xml` file for the tests, this should (most-likely) be under the `strongbox-common-resources/src/test/resources`.
> Yes , it should be but as I said it was not working from `strongbox-common-resources/src/test/resources` (may be due to test-jar) , so put it in `strongbox-common-resources/src/main/resources` kind of brute force :)
> I don't like the concept of having multiple exact copies of the same resources across different modules. This is makes it harder to maintain the same content in them on the long run.
> Agreed, as I said `strongbox-client` is not dependent on `testing-core` so created a new separate `logback-test.xml` for client module. Will try with `testing-core` dependency in `strongbox-client` so that it can be removed.
> Also, why do we need to include `org/springframework/boot/logging/logback/base.xml`? We should try to have rules for the packages what we have defined in the `logback-spring.xml`. Of course, they can be at a different logging level when doing development.
> Sure , will try suggested change.

anki2189:
> First of all, almost all of our modules are configured to produce a `test-jar` (containing their code from `${basedir}/test/classes` (this also included the resource files from `${basedir}/src/test/resources`)). This is done by adding the following to your `pom.xml`:
> 
> ```
>             <plugin>
>                 <groupId>org.apache.maven.plugins</groupId>
>                 <artifactId>maven-jar-plugin</artifactId>
>                 <executions>
>                     <execution>
>                         <goals>
>                             <goal>test-jar</goal>
>                         </goals>
>                     </execution>
>                 </executions>
>             </plugin>
> ```
> 
> (I can see that this has not been added to the `strongbox-common-resources` module, which is most-likely, why it didn't work for you, as you told me the other day in the chat).

It was miss from my side as I am new to this project and haven't explored how this is working in other modules. I will try and test this again with suggested changes.

anki2189:
> Thanks for your pull request, @anki2189 !
> 
> I'm afraid I disagree with with the introduction of both proposed `logback-test.xml` files.
> 
> First of all, almost all of our modules are configured to produce a `test-jar` (containing their code from `${basedir}/test/classes` (this also included the resource files from `${basedir}/src/test/resources`)). This is done by adding the following to your `pom.xml`:
> 
> ```
>             <plugin>
>                 <groupId>org.apache.maven.plugins</groupId>
>                 <artifactId>maven-jar-plugin</artifactId>
>                 <executions>
>                     <execution>
>                         <goals>
>                             <goal>test-jar</goal>
>                         </goals>
>                     </execution>
>                 </executions>
>             </plugin>
> ```
> 
> (I can see that this has not been added to the `strongbox-common-resources` module, which is most-likely, why it didn't work for you, as you told me the other day in the chat).
> 

It was miss from my side as I am new to this project and haven't explored how this is working in other modules. I will try and test this again with suggested changes.

> Secondly, if we need a `logback-test.xml` file for the tests, this should (most-likely) be under the `strongbox-common-resources/src/test/resources`.
> 
Yes , it should be but as I said it was not working from `strongbox-common-resources/src/test/resources` (may be due to test-jar) , so kept it in `strongbox-common-resources/src/main/resources` kind of brute force :)

> I don't like the concept of having multiple exact copies of the same resources across different modules. This is makes it harder to maintain the same content in them on the long run.
> 

Agreed, as I said `strongbox-client` is not dependent on `testing-core` so created a new separate `logback-test.xml` for client module. Will try to include `testing-core` dependency in `strongbox-client` so that it can be removed.

> Also, why do we need to include `org/springframework/boot/logging/logback/base.xml`? We should try to have rules for the packages what we have defined in the `logback-spring.xml`. Of course, they can be at a different logging level when doing development.

Sure , will try suggested change.

strongbox-resources/strongbox-common-resources/pom.xml
```diff
@@ -25,14 +25,26 @@
...
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-source-plugin</artifactId>
-            </plugin>
-        </plugins>
-    </build>
+	<build>
```
carlspring: Cosmetic: Coul you please use the same indentation that we do? We use four spaces, instead of tabs. Thanks! :)

anki2189: Done

carlspring: Thanks! :)

steve-todorov:
Hi @anki2189, could you please `rebase` and checkout the failing test case? :)

@carlspring could you re-review after the rebase so we can merge afterwards?

anki2189:
> Hi @anki2189, could you please `rebase` and checkout the failing test case? :)
> 
> @carlspring could you re-review after the rebase so we can merge afterwards?

Rebased with master.

steve-todorov:
Thanks! This has reduced our Jenkins logs from `~49MB` to `~28MB` which is a huge improvement! Many thanks! :)

// cc @carlspring

strongbox-resources/strongbox-common-resources/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <include resource="org/springframework/boot/logging/logback/base.xml" />
```
carlspring: @anki2189 : I think I asked this in the past and I'm not sure I got an answer to it: why are you including this `base.xml`? If anything needs to be included, it should be the `logback-spring.xml` and we should be overriding the logger levels from there.

@sbespalov : What are your thoughts?

sbespalov: it’s not entirely clear to me why we need this `logback-test.xml` in general. @anki2189 can you please clarify why you think we need it?

anki2189: @sbespalov : logback-test.xml is created only to be used by test cases. I tried to include logger in `logback-spring.xml` but I am clueless why it is not working.

@carlspring : `base.xml` has to be included for importing appenders. If we include `logback-spring.xml` it will will but with errors and warnings. 
It may be related to this .https://github.com/spring-projects/spring-boot/issues/7955

sbespalov: > I tried to include logger in logback-spring.xml but I am clueless why it is not working.

what you trying to include there? for now it's seems that there is no reasons for another logback configuration only for tests.

anki2189: > > I tried to include logger in logback-spring.xml but I am clueless why it is not working.
> 
> what you trying to include there? for now it's seems that there is no reasons for another logback configuration only for tests.

I want to change log level for spring test package `org.springframework.test`

sbespalov: it should be possible to do this in `logback-spring.xml ` directly

anki2189: No it not working in `logback-spring.xml`

sbespalov: you mean that `<logger name="org.springframework.test" level="WARN"/>` not working in `logback-spring.xml`?

anki2189: Yes, `<logger name="org.springframework.test" level="WARN"/> `not working in `logback-spring.xml`

PR #1612, opened 2019-11-28
Issue 841 - Create a Debian distribution package

### LINKED ISSUE #841: Create a Debian distribution
# Task Description

We are currently producing `zip` and `gzip` packages for our distribution. We need to be able to create `deb` distributions.

We need to evaluate the following plugins (please, feel free to add any others you might think would be a good fit) and decide which one we should use:
* [jdeb](https://github.com/tcurdt/jdeb) : This plugin allows you to build the project on any OS without any additional binaries. We need to check, if either of the other two allow building without additional binaries. The project seems active on Github. So far, this project seems to be our first choice.
* [io.solit.maven:deb-maven-plugin](https://github.com/io-solit/deb-maven-plugin) : The project seems active on Github.
* [net.sf.debian-maven:debian-maven-plugin](http://debian-maven.sourceforge.net/usage.html) : This project seems to have been inactive over the past four years. (Perhaps we should consider the other two options first).

The place where we produce our assemblies is the [`strongbox-distribution`](https://github.com/strongbox/strongbox/tree/master/strongbox-distribution) module.

# Task Relationships

This task:
* Depends on: ~#801~
* Relates to: #840

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @fuss86

TheCakeIsNaOH:
# Pull Request Description

This pull request closes #841 

# Acceptance Test

* [x] N/A Running `mvn spring-boot:run` in the `strongbox-web-core` still starts up the application correctly.
* [x] N/A Building the code and running the `strongbox-distribution` from a `zip` or `tar.gz` still works.
* [x] N/A The tests in the [`strongbox-web-integration-tests`](https://github.com/strongbox/strongbox-web-integration-tests/) still run properly.

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes, please see #...
  * [x] No

  * [x] Yes, https://github.com/strongbox/strongbox-docs/pull/70
  * [ ] No

# Info

* This does not totally uninstall cleanly even with a purge, it leaves strongbox-vault. I thought that it is best to not remove the vault uninstall/purge. This is up for discussion. 
* This does not remove the strongbox user and group upon uninstall as per [Debian recommendation](https://wiki.debian.org/AccountHandlingInMaintainerScripts)
* It takes a dependency on Java 8 runtime or java 8 runtime headless. These are provided by the JRE packages on Debian, and the JRE is installed as a dependency on the JDK. 
* This is NOT ready to put into an official repository, there are a couple of things like adding GPG signing and more maintainer info that are required by Debian. 
* Should be tested in more scenarios, VM, docker, native, Ubuntu spins, Mint, Pop-OS, others?
* This will build the deb on both Windows and on Linux

# TODO
* [x] wiki changes and make PR
* [x] figure out if strongbox-vault should be removed upon package purge
* [x] tested on Ubuntu 18.04 and 19.10 on docker, along with a Debian 10 VM

### DIFF
--- a/strongbox-distribution/src/main/resources/deb/control
+++ b/strongbox-distribution/src/main/resources/deb/control
@@ -4,6 +4,6 @@ Architecture: all
 Section: Miscellaneous
 Priority: low
 Homepage: https://github.com/strongbox/strongbox
-Description: Modern artifact repository manager
-Maintainer: Carlspring <carlspring@gmail.com>
+Description: The Strongbox artifact repository manager
+Maintainer: Martin Todorov <carlspring@gmail.com>
 Depends: java8-runtime | java8-runtime-headless
--- a/strongbox-distribution/src/main/resources/deb/control
+++ b/strongbox-distribution/src/main/resources/deb/control
@@ -1,0 +1,8 @@
+Package: strongbox
+Version: [[version]]
+Architecture: all
+Section: Miscellaneous
+Priority: low
+Homepage: https://github.com/strongbox/strongbox
+Description: Modern artifact repository manager
+Maintainer: Carlspring <carlspring@gmail.com>
--- a/strongbox-distribution/src/main/resources/deb/control
+++ b/strongbox-distribution/src/main/resources/deb/control
@@ -6,3 +6,4 @@ Priority: low
 Homepage: https://github.com/strongbox/strongbox
 Description: Modern artifact repository manager
 Maintainer: Carlspring <carlspring@gmail.com>
+Depends: java8-runtime | java8-runtime-headless
--- a/strongbox-distribution/src/main/resources/deb/control
+++ b/strongbox-distribution/src/main/resources/deb/control
@@ -6,4 +6,4 @@ Priority: low
 Homepage: https://github.com/strongbox/strongbox
 Description: The Strongbox artifact repository manager
 Maintainer: Martin Todorov <carlspring@gmail.com>
-Depends: java8-runtime | java8-runtime-headless
+Recommends: java8-runtime | java8-runtime-headless
--- a/strongbox-distribution/src/main/resources/deb/postinst
+++ b/strongbox-distribution/src/main/resources/deb/postinst
@@ -5,5 +5,5 @@ mkdir -p /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault
 getent group strongbox >/dev/null 2>&1 || groupadd --system strongbox
 id -u strongbox >/dev/null 2>&1 || adduser --quiet --system --ingroup strongbox --home /opt/strongbox strongbox
 
-chown -R strongbox:strongbox /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault
-chmod -R 770 /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault
+chown -R strongbox:strongbox /opt/strongbox /opt/strongbox-vault
+chmod -R 770 /opt/strongbox /opt/strongbox-vault
--- a/strongbox-distribution/src/main/resources/deb/postinst
+++ b/strongbox-distribution/src/main/resources/deb/postinst
@@ -1,6 +1,9 @@
 #!/bin/sh
-mkdir /opt/strongbox-vault
-groupadd strongbox
-useradd -d /opt/strongbox -g strongbox --system strongbox
-chown -R strongbox:strongbox /opt/strongbox
-chown -R strongbox:strongbox /opt/strongbox-vault
+set -e
+mkdir -p /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault 
+
+getent group strongbox >/dev/null 2>&1 || groupadd --system strongbox
+id -u strongbox >/dev/null 2>&1 || adduser --quiet --system --ingroup strongbox --home /opt/strongbox strongbox
+
+chown -R strongbox:strongbox /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault
+chmod -R 770 /opt/strongbox/tmp /opt/strongbox/logs /opt/strongbox-vault
--- a/strongbox-distribution/src/main/resources/deb/postinst
+++ b/strongbox-distribution/src/main/resources/deb/postinst
@@ -1,0 +1,6 @@
+#!/bin/sh
+mkdir /opt/strongbox-vault
+groupadd strongbox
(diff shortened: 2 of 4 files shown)


carlspring:
Hi @TheCakeIsNaOH ,

Thank you very much for this fantastic contribution!

You will also have to update [this section](https://github.com/strongbox/strongbox/blob/master/strongbox-distribution/pom.xml#L550) to also include the `deb` package. Could you please also add a link to the `deb` package in the `README.md`?

Thanks! :)

TheCakeIsNaOH:
Done.

strongbox-distribution/src/main/resources/deb/postinst
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -e
```
steve-todorov: This would not cover any `pipe fails` (which could happen at lines `5` and `6`). 
To cover those cases you'd need to do `set -euo pipefail`. :)

strongbox-distribution/src/main/resources/deb/control
```diff
@@ -0,0 +1,9 @@
+Package: strongbox
+Version: [[version]]
+Architecture: all
+Section: Miscellaneous
+Priority: low
+Homepage: https://github.com/strongbox/strongbox
+Description: Modern artifact repository manager
+Maintainer: Carlspring <carlspring@gmail.com>
+Depends: java8-runtime | java8-runtime-headless
```
steve-todorov: I think we probably shouldn't enforce these dependencies but rather have them as [`Suggests`](https://linux.die.net/man/5/deb-control) so people could opt-in to install them via `apt install --install-suggests`. 
There are valid cases where people already have a version of `JRE` installed on the server and enforcing the dependency could lead to unwanted upgrades and issues.

// @carlspring thoughts?

strongbox-distribution/src/main/resources/deb/control
```diff
@@ -0,0 +1,9 @@
+Package: strongbox
+Version: [[version]]
+Architecture: all
+Section: Miscellaneous
+Priority: low
+Homepage: https://github.com/strongbox/strongbox
+Description: Modern artifact repository manager
+Maintainer: Carlspring <carlspring@gmail.com>
```
carlspring:
```suggestion
Maintainer: Martin Todorov <carlspring@gmail.com>
```

strongbox-distribution/src/main/resources/deb/control
```diff
@@ -0,0 +1,9 @@
+Package: strongbox
+Version: [[version]]
+Architecture: all
+Section: Miscellaneous
+Priority: low
+Homepage: https://github.com/strongbox/strongbox
+Description: Modern artifact repository manager
```
carlspring:
```suggestion
Description: The Strongbox artifact repository manager
```

strongbox-distribution/src/main/resources/deb/control
carlspring: Tough one... On the one hand -- you need Java installed, on the other -- you don't want to force a specific version of the JDK, as long as it's a recent 1.8 one.

Let's go for the `Suggests`.

carlspring:
@ptirador , @v3rtumnus , 

Would you like to try this out?

strongbox-distribution/src/main/resources/deb/control
TheCakeIsNaOH: Done

strongbox-distribution/src/main/resources/deb/control
TheCakeIsNaOH: Done

strongbox-distribution/src/main/resources/deb/postinst
TheCakeIsNaOH: sh is the shell that is supposed to be used for this, and pipefail is not available in sh. 

I will get back to this.

strongbox-distribution/src/main/resources/deb/control
TheCakeIsNaOH: The JDK deb packages I looked at all took a dependency on the JRE, and the JREs all had `Provides` on the java8-runtime, headless or both, so I am pretty sure 

> There are valid cases where people already have a version of JRE installed on the server and enforcing the dependency could lead to unwanted upgrades and issues.

Can you describe in more detail what you mean here?

> you don't want to force a specific version of the JDK, as long as it's a recent 1.8 one.

I don't think this does, as long as the Provides field is filled in, which it should be.

strongbox-distribution/src/main/resources/deb/postinst
steve-todorov: Oh, you're right - my bad. I don't remember if `/bin/bash` was available on all Debian/Ubuntu based distributions by default so maybe we should leave it as it is then. If you happen to find any way to make it quit at the first error it gets that would be good but otherwise - we can just merge it as it is for now.

Just FYI - [this](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/) is how you'd do it in `/bin/bash`.

strongbox-distribution/src/main/resources/deb/control
steve-todorov: > Can you describe in more detail what you mean here?

By using `Depends` you are making `apt` require the `java8-runtime` package to be installed or otherwise Strongbox won't be installed. However, there are many ways to setup a server. You could extract a `JDK/JRE` in a specific directory (i.e. you've downloaded from `openjdk` or `oraclejdk`) and exported `JAVA_HOME` -- or you could have installed it via `apt install`. If you already have a custom JDK/JRE setup - then installing `java-runtime8` would make no sense. 

After having another look at the official docs, [`Recommends`](https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends) is a better choice (rather than my initial suggestion to use `Suggests`). This will install the the dependency by default when you do `apt install ./strongbox-distribution-1.0-SNAPSHOT.deb` but would allow the flexibility to add skip the jre dependency with when doing `apt --no-install-recommends install ./strongbox-distribution-1.0-SNAPSHOT.deb` (should it be necessary)

steve-todorov:
I've tried installing the debian package on ubuntu 19.10 and it works as expected. :)

strongbox-distribution/src/main/resources/deb/control
TheCakeIsNaOH: Right, I had not considered about non-deb java installs.

carlspring:
Thank you for your hard work, guys!

PR #1642, opened 2020-01-14
Implementation for Raw Artifacts with specified size

anki2189:
# Pull Request Description

This pull request closes #1540 

# Acceptance Test

* [x] Building the code with `mvn clean install -Dintegration.tests` still works.
* [x] Running `mvn spring-boot:run` in the `strongbox-web-core` still starts up the application correctly.
* [x] Building the code and running the `strongbox-distribution` from a `zip` or `tar.gz` still works.
* [x] The tests in the [`strongbox-web-integration-tests`](https://github.com/strongbox/strongbox-web-integration-tests/) still run properly.

# Questions

* Does this pull request break backward compatibility? 
  * [ ] Yes
  * [x] No

* Does this pull request require other pull requests to be merged first? 
  * [ ] Yes, please see #...
  * [x] No

* Does this require an update of the documentation?
  * [ ] Yes, please see strongbox/strongbox-docs#{PR_NUMBER}
  * [x] No

PR #1646, opened 2020-01-19
Reverting #1642 "Implementation for Raw Artifacts with specified size"

steve-todorov:
Reverts strongbox/strongbox#1642 which seems to have broken the master.


steve-todorov:
After discussing this with @anki2189 and @carlspring this revert won't fix the master build failures. The reason for the fails is because maven central has rushed into switching to `https` without notice (actually having one, but during Christmas and the New Year holidays so nobody actually noticed it):
```
501 HTTPS Required. 
Use https://repo.maven.apache.org/maven2/
More information at https://links.sonatype.com/central/501-https-required
```
We'll need to fix the failing test cases by switching them to use `https` instead of `http`.

PR #1768, opened 2020-05-12
Issue 1511 : Fix credentials shown in logs

### LINKED ISSUE #1511: Make sure that no passwords are shown in the logs
# Task Description

When a user authenticates, the logs may contain his hashed password. We should make sure that we don't leak any credential information in the logs (even hashed passwords).

# Tasks

* [ ] Enumerate all possible classes that uses credentials / passwords  
* [ ] Make sure their `toString` methods do not contain sensitive data
* [ ] Make sure  credentials / passwords are not logged anywhere

# Help

* Points of contact:
  * @carlspring
  * @fuss86
  * @sbespalov
  * @steve-todorov

hT013:
# Pull Request Description

This pull request closes #1511 

# Acceptance Test

# Questions

  * [ ] Yes
  * [x] No

  * [ ] Yes
  * [x] No

  * [x] No

### DIFF
--- a/strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/configuration/MutableProxyConfiguration.java
+++ b/strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/configuration/MutableProxyConfiguration.java
@@ -144,7 +144,6 @@ public class MutableProxyConfiguration
                           .add("host", host)
                           .add("port", port)
                           .add("username", username)
-                          .add("password", password)
                           .add("type", type)
                           .add("nonProxyHosts", nonProxyHosts)
                           .toString();
--- a/strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
+++ b/strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
@@ -138,7 +138,7 @@ public class WebConfig
 
         };
         result.setIncludeQueryString(true);
-        result.setIncludeHeaders(true);
+        result.setIncludeHeaders(false);
         result.setIncludeClientInfo(true);
         return result;
     }
--- a/strongbox-web-core/src/main/java/org/carlspring/strongbox/security/authentication/JwtTokenFetcher.java
+++ b/strongbox-web-core/src/main/java/org/carlspring/strongbox/security/authentication/JwtTokenFetcher.java
@@ -38,7 +38,7 @@ public interface JwtTokenFetcher
             if (headerMatcher.matches())
             {
                 String token = headerMatcher.group(1);
-                logger.debug("Bearer Authorization header found with token {}", token);
+                logger.debug("Bearer Authorization header found");
                 return Optional.of(token);
             }
         }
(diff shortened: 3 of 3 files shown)


strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/configuration/MutableProxyConfiguration.java
```diff
@@ -144,7 +144,6 @@ public String toString()
                           .add("host", host)
                           .add("port", port)
                           .add("username", username)
-                          .add("password", password)
```
steve-todorov, line 147: This should not be removed, but rather displayed as redacted:
```
    .add("password", "[credentials redacted]")
```

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
```diff
@@ -138,7 +138,7 @@ protected String createMessage(HttpServletRequest request,
 
         };
         result.setIncludeQueryString(true);
-        result.setIncludeHeaders(true);
```
steve-todorov, line 141: Not sure if this is a good idea since this will remove the headers. We should probably add `beforeRequest` and `afterRequest` and redact the commonly known sensitive headers (i.e. `Authentication: Bearer TOKEN` => `Authentication: Bearer [credentials redacted]`). 

@sbespalov thoughts?

strongbox-storage/strongbox-storage-core/src/main/java/org/carlspring/strongbox/configuration/MutableProxyConfiguration.java
carlspring: Excuse me for joining in late, but two things:
1) If the password is null, append `null`. (This should never be null, BUT it might as well be the case that during some test, someone might have overlooked it. Having it printed out in the logs as `null` should not be a security concern, in my opinion).
2) If the passwrod is set, then I think the better option would be to mask it with `********`. (People will know it's masked/redacted, for sure).

strongbox-web-core/src/main/java/org/carlspring/strongbox/config/WebConfig.java
hT013: Added `result.setHeaderPredicate(s -> !(s.equals("Authorization") || s.equals("Cookie")));` which mask the authorization and cookies in header.

Header shown in log  -> `[method=GET;GET /api/account, client=0:0:0:0:0:0:0:1, user=admin, headers=[Authorization:"masked", Cookie:"masked", Accept:"application/json", X-Requested-With:"XMLHttpRequest", User-Agent:"Mozilla/5.0 (X11; Linux x86_64; rv:76.0) Gecko/20100101 Firefox/76.0", Connection:"keep-alive", Referer:"http://localhost:48080/my-account", Host:"localhost:48080", Accept-Language:"en-US,en;q=0.5", Accept-Encoding:"gzip, deflate", DNT:"1"]]`

PR #1793, opened 2020-06-16
issues/1790: Make it possible to use standalone cassandra

### LINKED ISSUE #1790: Make it possible to use standalone cassandra
# Task Description

We need to allow usage of standalone Cassandra instances for cases where you'd like to run a cluster of Cassandra instances.
There should be an environment variable called `STRONGBOX_REMOTE_CASSANDRA` and a `strongbox.yaml` configuration variable which would need to be an array of strings as shown in JG's docs ([storage.hostname](https://docs.janusgraph.org/basics/configuration-reference/#storage))

`strongbox.yaml`: 
```yaml
strongbox: 
  configuration:
    database:
      # If true -> starts embedded cassandra; 
      # if false -> passes the string array below to JG
      embedded: true|false
      # This term comes from datastax's driver and refers to the coordinator node you will be connecting to
      # If embedded:true, this option should be optional and should default to 127.0.0.1:9042
      contactPoints: [ "1.2.3.4:9042", "2.3.4.5:9042" ]
      authentication: (optional, if missing assume `username=null` and `password=`null`)
        username: "my-cassandra-user" (if blank, `username=null`)
        password: "my-cassandra-password" (if blank, `password=null`)
```

Maybe we could have a `janusgraph:` section here which directly maps values to `JanusGraphFactory.build().set(key,val)` as well, but if it's too much work we can add a follow-up 

`STRONGBOX_REMOTE_CASSANDRA`:
* `STRONGBOX_REMOTE_CASSANDRA=1.2.3.4:9042,2.3.4.5:9042` should be translated as `contactPoints: [1.2.3.4:9042, 2.3.4.5:9042]`
* `STRONGBOX_REMOTE_CASSANDRA_USER=username` -> `authentication.username=username` 
* `STRONGBOX_REMOTE_CASSANDRA_PASS=password` -> `authentication.password=password`

# Task Relationships

This task:
* Is a sub-task of: #1649

sbespalov:
# Pull Request Description

This pull request closes #1790

# Acceptance Test

# Questions

  * [ ] Yes
  * [ ] No

  * [ ] Yes, please see #...
  * [ ] No

  * [ ] No


strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/EmbeddedDbServerConfiguration.java
```diff
@@ -57,13 +59,13 @@ EmbeddedDbServer embeddedDbServer(CassandraEmbeddedConfiguration cassandraConfig
             logger.info(String.format("Sotorage extracted to [%s].", cassandraConfiguration.getStorageFolder()));
```
carlspring, line 59: Typo: `Sotorage` -->` `Storage`.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/EmbeddedDbServerConfiguration.java
```diff
@@ -133,9 +130,10 @@ JanusGraphConfiguration janusGraphConfiguration()
     }
 
     @Bean
-    CassandraEmbeddedConfiguration cassandraEmbeddedConfiguration(JanusGraphConfiguration janusGraphConfiguration)
+    CassandraEmbeddedConfiguration cassandraEmbeddedConfiguration(@Value("${strongbox.db.janus-graph.storage-root}") String storageRoot,
```
carlspring, line 133: This is a very strange name for a property. Could you please rename it to `strongbox.db.janusgraph.storage.root`? Thanks!

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/RemoteDbServerConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Bean
+    JanusGraph JanusGraph(JanusGraphServer server)
+        throws Exception
+    {
+        return new StrongboxSchema().createSchema(server.getJanusGraph());
+    }
+
+    @Bean
+    @ConfigurationProperties(prefix = "strongbox.db.janus-graph")
```
carlspring, line 41: Could you please change this to `strongbox.db.janusgraph`

strongbox-data-service/src/main/resources/etc/conf/janusgraph_EMBEDDED.yaml
```diff
@@ -0,0 +1,7 @@
+strongbox.db:
+    janus-graph:
```
carlspring, line 2: I propose changing this to `janusgraph` in line with the above.

strongbox-data-service/src/main/resources/etc/conf/janusgraph_EMBEDDED.yaml
```diff
@@ -0,0 +1,7 @@
...
+        storage-username: cassandra
+        storage-password: cassandra
+        storage-host: 127.0.0.1
+        storage-port: 49142
+        storage-root: ${strongbox.vault}/db
```
carlspring, line 7: Could all of these be `storage.xyz`, please? Thanks!

strongbox-data-service/src/main/resources/etc/conf/janusgraph_REMOTE.yaml
```diff
@@ -0,0 +1,6 @@
+strongbox.db:
+    janus-graph:
+        storage-username: cassandra
+        storage-password: cassandra
+        storage-host: 127.0.0.1
+        storage-port: 49142
```
carlspring, line 6: As above.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/EmbeddedDbServerConfiguration.java
```diff
@@ -46,7 +48,7 @@
     public static final String PATH_STRONGBOX_DB = "META-INF/org/carlsparing/strongbox/db";
 
     @Bean
-    EmbeddedDbServer embeddedDbServer(CassandraEmbeddedConfiguration cassandraConfiguration,
+    JanusGraphServer embeddedDbServer(CassandraEmbeddedConfiguration cassandraConfiguration,
```
carlspring, line 51: Shouldn't this rather be `remoteDbServer`?

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/InMemoryDbServerConfiguration.java
```diff
@@ -25,27 +23,20 @@
 public class InMemoryDbServerConfiguration implements Condition
 {
 
-    private static final Logger logger = LoggerFactory.getLogger(InMemoryDbServerConfiguration.class);
-
     @Bean
-    EmbeddedDbServer embeddedDbServer(JanusGraphConfiguration janusGraphConfiguration)
+    JanusGraphServer embeddedDbServer(JanusGraphConfiguration janusGraphConfiguration)
     {
```
carlspring, line 29: You can remove this empty line.

strongbox-data-service/src/main/resources/etc/conf/janusgraph_EMBEDDED.yaml
sbespalov: this is the naming convention that Spring using to map it on property names. So `storage-username` will be mapped to `JanusGraphProperties.storageUsername`.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/EmbeddedDbServerConfiguration.java
sbespalov: This name is for consistency between all the `...ServerConfiguration` classes. It will be better to have same bean name in all configurations.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/InMemoryDbServerConfiguration.java
sbespalov: done

strongbox-web-core/src/main/java/ru/sbespalov/test/Game.java
```diff
@@ -0,0 +1,51 @@
+package ru.sbespalov.test;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class Game
```
steve-todorov, line 10: @sbespalov it looks like this is not supposed to be here or am I missing something? :)

strongbox-data-service/src/main/resources/etc/conf/db_EMBEDDED.yaml
```diff
@@ -0,0 +1,7 @@
+strongbox:
+    db:
+      janusgraph:
+        config-location: ${strongbox.home}/etc/conf/janusgraph-cassandra.properties
+      cassandra:
+        config-locatoion: file:${strongbox.home}/etc/conf/cassandra.yaml
```
steve-todorov, line 6: There is a typo here:
```suggestion
        config-location: file:${strongbox.home}/etc/conf/cassandra.yaml
```

Also, if I were a user, I'd expect all of the `config-location` fields to work the same way and would not set the `file:` prefix. As far as I remember, the CassandraDaemon class allowed you to use `classpath:`, but I can't think of a use case right now in which we'd be using `classpath:` instead of `file:`? 

I think we should automatically add the `file:` prefix to the string in the backend part and if in the future we need to - we can make it more robust and allow for `classpath:`. What do you think?

strongbox-data-service/src/main/resources/etc/conf/db_REMOTE.yaml
```diff
@@ -0,0 +1,4 @@
+strongbox:
+    db:
+      janusgraph:
+        config-location: ${strongbox.home}/etc/conf/janusgraph-cassandra.properties
```
steve-todorov, line 4:
```suggestion
      janusgraph:
        # Note: Configuring JG with a remote Cassandra instance would require you to set the `storage.hostname` property in JanusGraph.
        config-location: ${strongbox.home}/etc/conf/janusgraph-cassandra.properties
```

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/RemoteDbServerConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public boolean matches(ConditionContext conditionContext,
+                           AnnotatedTypeMetadata metadata)
+
+    {
+        JanusGraphDbProfile profile = JanusGraphDbProfile.resolveProfile((ConfigurableEnvironment) conditionContext.getEnvironment());
+
+        return profile.getName().equals(JanusGraphDbProfile.PROFILE_REMOTE);
```
steve-todorov, line 47: The profile matching logic is very limiting because it assumes you will use **exactly** the `db_REMOTE.yaml` file for remote Cassandra configurations. If, like in my case, you have configured multiple clusters with different settings for testing and validation purposes you will have to go to `strongbox-db`, configure the file manually, wait a build for 1 minute, go back to the Strongbox project, build for another minute or two and only then proceed with testing. If you need to check the other cluster - you need to go back and do the same operation again which looses a lot of time. :) 

For now lets leave it as it is but if you have ideas on how to improve this - it would be an awesome follow-up task and we should discuss it in the chat.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/EmbeddedDbServerConfiguration.java
sbespalov: fixed

strongbox-data-service/src/main/resources/etc/conf/db_EMBEDDED.yaml
sbespalov: I thought that with protocol prefix it would be more flexible, so as you mentioned it could be `classpath` or `jar` protocols. For example initially I've used `jar` protocol at `strongbox-db-import` project. Probably it would be better to leave this "feature", so maybe you will decide to use `http` in the future :)

sbespalov: BTW, the `classpath` is not possible, the only similar option is `jar`.

strongbox-data-service/src/main/java/org/carlspring/strongbox/config/janusgraph/RemoteDbServerConfiguration.java
sbespalov: as an option you could override the `config-location` property with JDK options, like common Spring Environment variable. The profiles itself needed to have proper Spring context configuration, so for example with embedded profile we have cassandra instance as Spring bean.

strongbox-web-core/src/main/java/ru/sbespalov/test/Game.java
sbespalov: removed :)

strongbox-data-service/src/main/resources/etc/conf/db_EMBEDDED.yaml
sbespalov: `janusgraph.config-location` is now accept URL as agreed.

PR #1803, opened 2020-07-05
Limit users on UI using skip and limit

### LINKED ISSUE #1796: Rework `UserController.getUsers` method so that it takes `skip` and `limit` parameters
# Task Description

Current implementation fetches all users from DB, which wont work well in real life. We should be able to paginate users on UI with `skip` and `limit` parameters.

# Tasks

* [ ] Create `UserRepository.findUsers(Long skip, Integer limit)` method (see `ArtifactIdGroupQueries.findArtifacts` as example)
* [ ] Rework  `UserController.getUsers` to use `UserRepository.findUsers` with `skip` and `limit` parameters
* [ ] Provide Unit Tests for `UserController.getUsers`

# Task Relationships

This task:
* Is a sub-task of: #1649 

# Help

* Points of contact:
  * @carlspring
  * @sbespalov
  * @steve-todorov

anki2189:
# Pull Request Description

This pull request closes #1796 

# Acceptance Test

# Questions

  * [ ] Yes
  * [ ] No

  * [ ] Yes, please see #...
  * [ ] No

  * [ ] No


anki2189:
@sbespalov : Please review the changes

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/users/UserController.java
```diff
@@ -106,9 +106,10 @@
     @PreAuthorize("hasAuthority('VIEW_USER')")
     @GetMapping(produces = { MediaType.APPLICATION_JSON_VALUE })
     @ResponseBody
-    public ResponseEntity getUsers()
+    public ResponseEntity getUsers(@RequestParam(value = "skip") Long skip,
```
sbespalov, line 109: @steve-todorov this will require `skip` and `limit` parameters support on UI.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/environment/EnvironmentInfoController.java
```diff
@@ -85,14 +85,13 @@ public ResponseEntity getEnvironmentInfo()
 
         strongboxInfo.add(new EnvironmentInfo("storages", String.valueOf(storagesCount)));
 
-        Long usersCount = userService.getUsers()
+        Long usersCount = userService.getUsers(null, null)
```
sbespalov, line 88: @anki2189 such an implementation is not correct. On the other hand, the real user count will run slowly, so I suggest to drop `usersCount` from `strongboxInfo`. @steve-todorov will you mind?

strongbox-security/strongbox-user-management/src/main/java/org/carlspring/strongbox/users/service/impl/InMemoryUserService.java
```diff
@@ -41,7 +41,8 @@
     private SecurityTokenProvider tokenProvider;
 
     @Override
-    public Users getUsers()
+    public Users getUsers(Long skip,
```
sbespalov, line 44: @anki2189 we need `skip` and `limit` support here as well. Could you please also make this parameters required?

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/environment/EnvironmentInfoController.java
steve-todorov: I agree the implementation is wrong - it was meant to count the `users` from the `yaml` configuration file and will not work in the current context. Right now, I suggest we always return `0` and add a `TODO: Implement a user counter` comment above. In the long run we are probably going to rely on this information for licensing purposes.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/users/UserController.java
```diff
@@ -106,9 +106,10 @@
...
+                                   @RequestParam(value = "limit") Integer limit)
```
steve-todorov, line 110: Thanks for mentioning me @sbespalov! 

This is actually a breaking change, because it requires the request parameters to be present (which they are not in the UI). Lets make them optional for the `API` and leave some reasonable defaults. This will make it more accessible for people who are playing around with the API and will preserve backwards compatibility.

```suggestion
    public ResponseEntity getUsers(@RequestParam(value = "skip", required = false, defaultValue = 0) Long skip,
                                   @RequestParam(value = "limit", required = false, defaultValue = 25) Integer limit)
```

A test for the `optional` values should be added as well (i.e. `/api/users`).

@anki2189 would you like to have a go with the UI which should have a? If not - we should add a follow-up task linked to this one to since I'm currently busy with another task and it will be a while until I can get to this.

strongbox-security/strongbox-user-management/src/main/java/org/carlspring/strongbox/users/service/impl/InMemoryUserService.java
anki2189: Fixed.

strongbox-web-core/src/main/java/org/carlspring/strongbox/controllers/environment/EnvironmentInfoController.java
anki2189: Fixed.

sbespalov:
@anki2189 Jenkins build seems failed. Could you please check?

carlspring:
@anki2189 , @sbespalov ,

What is the status of this task?

carlspring:
@anki2189 , @sbespalov ,

What is the status of this pull request (apart from it needing a rebase)?

anki2189:
> @anki2189 , @sbespalov ,
> 
> What is the status of this pull request (apart from it needing a rebase)?

It also require change in UI.

